PR #5, opened 2018-10-11
derive(Arbitrary)

### LINKED ISSUE #4: Auto-deriving Arbitrary
Similar to what I did for quickcheck, it should be easy enough to write a derive macro for `Arbitrary`, perhaps even with some shrinking support for simple types. Should I do it in another crate or send a PR your way to integrate it?

llogiq:
This adds a small macro for deriving Arbitrary. No shrinking specialization yet. Any ideas on what to do for shrinking?

This closes #4

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,0 +1,116 @@
+extern crate arbitrary;
+extern crate syn;
+#[macro_use]
+extern crate synstructure;
+extern crate proc_macro2;
+
+use proc_macro2::TokenStream;
+
+decl_derive!([Arbitrary] => arbitrary_derive);
+
+fn arbitrary_derive(s: synstructure::Structure) -> TokenStream {
+    if s.variants().len() == 1 { // struct
+        let con = s.variants()[0].construct(|_, _| quote! { Arbitrary::arbitrary(u)? });
+        s.gen_impl(quote! {
+            extern crate arbitrary;
+
+            use arbitrary::{Arbitrary, Unstructured};
+
+            gen impl Arbitrary for @Self {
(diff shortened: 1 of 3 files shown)


nagisa:
NB: shrinking is not really implemented for the manual types either, I wouldn’t worry about that too much I think.

derive/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        s.gen_impl(quote! {
+            extern crate arbitrary;
+
+            use arbitrary::{Arbitrary, Unstructured};
+
+            gen impl Arbitrary for @Self {
+                fn arbitrary<U: Unstructured + ?Sized>(u: &mut U) -> Result<Self, U::Error> {
+
+
+                    Ok(match (u64::from(<u32 as Arbitrary>::arbitrary(u)?) * #count) >> 32 {
```
nagisa: I’m having a hard time understanding what this means. We generate u32, convert it to u64, multiply by the number of variants and then drop the low bytes? What is the purpose of this complex logic? Why isn’t the genereated number taken modulo variant count?

Perhaps a comment would be appropriate.

llogiq: This version introduces less bias than modulo and is due to [lemire](https://lemire.me/blog/2016/06/30/fast-random-shuffling/).

nagisa: Please extend the comment with this blog post and also mention that this has a very slight bias.

nagisa:
Thanks!

llogiq:
You're welcome.

nagisa:
I’ve took liberty of [publishing the crate](https://crates.io/crates/derive_arbitrary)

Issue #8, opened 2019-11-26
Arbitrary impl for char is slow


Manishearth:
The current implementation, when it hits an invalid character, loops downward by one bit each time. This is slow, all of the invalid `char` values are between `0xD800` and  `0xDFFF`, or greater than `0x10ffff` (already dealt with with the mask)

This means that we spend a lot of time just looping here, especially when creating strings. We should instead just replace these with null or decrement the fourth nibble.

Manishearth:
Also, we might want to try directly constructing strings? In the past I've had good luck with just `String::from_utf8` on byte slices whilst fuzzing. It's fast and it has found many bugs. The current `String` implementation seems unable to find things even in cases where the testcase is literally "panic on this particular six-character string"

Manishearth:
Sadly the current design of the `Arbitrary` trait leaves no room for "try to construct a string, error out if it fails" :/

fitzgen:
> Sadly the current design of the `Arbitrary` trait leaves no room for "try to construct a string, error out if it fails" :/

I guess you could do a lossy utf8 conversion, but that might not always be appropriate.

nagisa:
Very open to larger changes to the trait, including adding potential error branches, but I personally think that people do not want to care about these details and would rather just get something.

Especially relevant if and when rngs get involved behind this trait.

Might be worth looking at how `quickcheck` has implemented generation of `char`.

Manishearth:
quickcheck is able to request infinite bytes, we cannot. It uses the `rand` crate's way of requesting chars with standard distribution (https://docs.rs/rand/0.7.2/src/rand/distributions/other.rs.html#42-61) but that does mean the input bytestring will not look at all like the output because it will be shifted, which isn't great for fuzzers -- fuzzers like to look at programs and generate things containing strings found in the program.

>  I personally think that users of this trait do not want to care about these details and would rather just get something.

The main user of this trait is the fuzzer stuff, and in fuzzing it makes more sense to reject early than to spend lots of effort trying to come up with a valid value.

Manishearth:
I'm not strong on having the reject-early stuff, but we should definitely fix it so that in the case where the fuzzer gives us an invalid byte we don't put lots of effort into "fixing" it

nagisa:
change the impl to be `for Option<char>` instead, maybe?

Manishearth:
Perhaps. We should fix the perf issue first, and then look at the API if
necessary

[...]

PR #14, opened 2019-12-24
Implement shrink for a bunch of stuff

fitzgen:
More on the way, but this felt like a good checkpoint.

r? @Manishearth

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -204,6 +204,14 @@ impl Arbitrary for char {
             }
         }
     }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let x = *self as u32;
+        Box::new(x.shrink().filter_map(|x| {
+            use std::convert::TryFrom;
+            char::try_from(x).ok()
+        }))
+    }
 }
 
 impl Arbitrary for AtomicBool {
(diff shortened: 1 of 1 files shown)


fitzgen:
Phew! Ok that's all of them :-p

Manishearth:
If you're doing a lot of work on this crate can you also fix https://github.com/rust-fuzz/rust_arbitrary/issues/8 ?

src/lib.rs
```diff
@@ -204,6 +204,14 @@ impl Arbitrary for char {
             }
         }
     }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let x = *self as u32;
+        Box::new(x.shrink().filter_map(|x| {
+            use std::convert::TryFrom;
+            char::try_from(x).ok()
```
Manishearth: this might be problematic for unicode correctness, it may abort early for large code points

fitzgen: Hm I’m not sure how I’m familiar enough with Unicode to write this method properly. Do you have a suggested implementation or should I skip this method and drop this commit from the pull?

Manishearth: Leave it for now, just make fixing this a part of #8

PR #16, opened 2019-12-24
Better impl for char

Manishearth:
Fixes https://github.com/rust-fuzz/rust_arbitrary/issues/8

The String impl is still not good enough, since it blindly calls `char::from_arbitrary`, which doesn't work as well. We should do something slightly more UTF8 aware.

r? @fitzgen

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -188,20 +188,18 @@ impl_arbitrary_for_floats! {
 
 impl Arbitrary for char {
     fn arbitrary<U: Unstructured + ?Sized>(u: &mut U) -> Result<Self, U::Error> {
-        const CHAR_MASK: u32 = 0x001f_ffff;
-        let mut c = <u32 as Arbitrary>::arbitrary(u)? & CHAR_MASK;
-        loop {
-            // Cannot do rejection sampling which the rand crate does, because it may result in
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -188,20 +188,17 @@ impl_arbitrary_for_floats! {
...
-            if let Some(c) = ::std::char::from_u32(c) {
-                return Ok(c);
-            } else {
-                c -= 1;
-            }
+        use ::std::char;
+        const CHAR_END: u32 = 0x0011_000;
+        // The size of the surrogate blocks
+        const SURROGATES_START: u32 = 0xD800;
+        let mut c = <u32 as Arbitrary>::arbitrary(u)? % CHAR_END;
```
fitzgen: I think this should do the same trick that the derive does do reduce bias in enum discriminants, no?

Manishearth: I don't want to shift it, because that changes valid inputs.

i want to keep valid inputs unchanged

Manishearth:
Hmm, another approach would be to parse as utf 8 instead, which might produce more success. That's trickier, though.

That would fix https://github.com/rust-fuzz/rust_arbitrary/issues/17 for free (aside from the container size)

Manishearth:
Landing this anyway, this approach is strictly better.

Issue #19, opened 2019-12-24
Tracking issue for 0.3.0 release


fitzgen:
So we have some nice improvements to this crate, and I think everything has technically been compatible with a 0.2.X release but I think it is a good time to revisit some of the public API and trait designs and make a 0.3.0 release. I figured it would be good to open an issue to talk about what we might want in it.

## TODO

* [x] Fill out public API doc comments, and add examples to each of them
* [x] Better `Arbitrary` for `String` implementation (#17)
* [x] Should `Arbitrary` have `Debug` as a super trait? (#7)
* [x] Should `Arbitrary` have `Clone` as a super trait? This would allow easier/better shrinking implementations (e.g. for various collections)
* [x] Refactor `Unstructured` to look more like [`FuzzedDataProvider`](https://github.com/rust-fuzz/libfuzzer-sys/blob/master/libfuzzer/utils/FuzzedDataProvider.h)? In particular, maybe we should get lengths from the end of the buffer like that does.
* [x] Add `Arbitrary::arbitrary_take_rest` methods to consume the rest of an `Unstructured`. (#18)
* [x] Add `Arbitrary::size_hint` and refactor `Unstructured::container_size` into `Unstructured::collection_len`, as discussed in #18.
* [x] Add a `fn arbitrary<A: Arbitrary>(&mut self) -> Result<A, Self::Error>` helper method to `Unstructured` to shorten invoking nested-arbitrary calls, so you can just do `let foo = u.arbitrary()?;` and let type inference figure things out for you?

As we surface more things to be done for 0.3.0 in the comments, I'll add them to this list, so that we have a single place for everything.

------------------------------

+cc @Manishearth: what do you think? Anything we should add to or remove from this list?

Manishearth:
The second last point is https://github.com/rust-fuzz/rust_arbitrary/issues/18

This seems about right, otherwise.

fitzgen:
I kind of think we should get rid of `RingBuffer` and only have a finite buffer (that potentially gives out zeros after its exhausted).

Because the values from a ring buffer can end up getting interpreted in multiple different ways, it is hard for the fuzzer to make a small tweak to the input without having that affect the generated structure in many different ways. To move as close to "small change in raw input leads to small change in generated structure" we should only use finite buffers.

Thoughts @Manishearth?

Manishearth:
Yeah, I don't think the ring buffer is useful. We can even make it so that it exits early if exhausted.

[...]

PR #20, opened 2019-12-27
A bunch of stuff

### LINKED ISSUE #10: Test failures on 32 bits arches
On both i686 and armv7hl, one test fails:

```
 /usr/bin/env CARGO_HOME=.cargo RUSTC_BOOTSTRAP=1 /usr/bin/cargo test -j6 -Z avoid-dev-deps --release --no-fail-fast
   Compiling arbitrary v0.2.0 (/builddir/build/BUILD/arbitrary-0.2.0)
     Running `/usr/bin/rustc --crate-name arbitrary src/lib.rs --color never --emit=dep-info,link -C opt-level=3 --test -C metadata=2b9ebc5100c56330 -C extra-filename=-2b9ebc5100c56330 --out-dir /builddir/build/BUILD/arbitrary-0.2.0/target/release/deps -L dependency=/builddir/build/BUILD/arbitrary-0.2.0/target/release/deps -Copt-level=3 -Cdebuginfo=2 -Clink-arg=-Wl,-z,relro,-z,now -Ccodegen-units=1 --cap-lints=warn`
    Finished release [optimized] target(s) in 5.04s
     Running `/builddir/build/BUILD/arbitrary-0.2.0/target/release/deps/arbitrary-2b9ebc5100c56330`
running 3 tests
test test::finite_buffer_fill_buffer ... ok
test test::ring_buffer_fill_buffer ... ok
test test::ring_buffer_container_size ... FAILED
failures:
---- test::ring_buffer_container_size stdout ----
thread 'test::ring_buffer_container_size' panicked at 'assertion failed: `(left == right)`
  left: `10`,
 right: `9`', src/lib.rs:596:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
failures:
    test::ring_buffer_container_size
test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

See build.log on https://koji.fedoraproject.org/koji/taskinfo?taskID=39452113

### LINKED ISSUE #18: "Fill all available space" solution for container size
Broadly, there are two kinds of things Arbitrary generates. Fixed-size objects that map to a fixed set of integers, and variable size objects, which can use something like `container_size()`.

Thing is, unlike, say, quickcheck, we already have a finite number of bytes as input. It would be nice if Arbitrary could be told "hey, I only have X bytes of data and you're my last consumer" as opposed to "slurp as much data as you want, I'll split the difference amongst the next consumers!"

This seems pretty tricky to design in the general case, but we could initially settle for types like `String` and `Vec` just filling directly.

fitzgen:
r? @Manishearth 

This is everything in #19 other than a better `Arbitrary for String` implementation, which I'm not super confident about implementing in a smart way. @Manishearth would you be up for doing that? Or outlining how you think that should be done?

Fixes #10

### DIFF
--- a/src/finite_buffer.rs
+++ b/src/finite_buffer.rs
@@ -1,0 +1,74 @@
+// Copyright © 2019 The Rust Fuzz Project Developers.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use crate::{Arbitrary, Unstructured};
+
+/// An enumeration of buffer creation errors
+#[derive(Debug, Clone, Copy)]
+pub enum BufferError {
+    /// The input buffer is empty, causing construction of some buffer types to
+    /// fail
+    EmptyInput,
+}
+
+/// A source of unstructured data with a finite size
+///
+/// This buffer is a finite source of unstructured data. Once the data is
+/// exhausted it stays exhausted.
+pub struct FiniteBuffer<'a> {
+    buffer: &'a [u8],
+    offset: usize,
+    max_len: usize,
+}
+
+impl<'a> FiniteBuffer<'a> {
+    /// Create a new FiniteBuffer
+    ///
+    /// If the passed `buffer` is shorter than max_len the total number of bytes
+    /// will be the bytes available in `buffer`. If `buffer` is longer than
+    /// `max_len` the buffer will be trimmed.
+    pub fn new(buffer: &'a [u8], max_len: usize) -> Result<Self, BufferError> {
+        let buf: &'a [u8] = if buffer.len() > max_len {
+            &buffer[..max_len]
+        } else {
+            // This branch is hit if buffer is shorter than max_len. We might
+            // choose to make this an error condition instead of, potentially,
+            // surprising folks with less bytes.
+            buffer
+        };
+
+        Ok(FiniteBuffer {
+            buffer: buf,
+            offset: 0,
+            max_len: buf.len(),
+        })
+    }
+}
+
+impl<'a> Unstructured for FiniteBuffer<'a> {
+    type Error = ();
+
+    fn fill_buffer(&mut self, buffer: &mut [u8]) -> Result<(), Self::Error> {
+        if (self.max_len - self.offset) >= buffer.len() {
+            let max = self.offset + buffer.len();
+            for (i, idx) in (self.offset..max).enumerate() {
+                buffer[i] = self.buffer[idx];
+            }
+            self.offset = max;
+            Ok(())
+        } else {
+            Err(())
+        }
+    }
+
+    // NOTE(blt) I'm not sure if this is the right definition. I don't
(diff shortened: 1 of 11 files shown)


src/finite_buffer.rs
```diff
@@ -0,0 +1,74 @@
...
+            }
+            self.offset = max;
+            Ok(())
+        } else {
+            Err(())
+        }
+    }
+
+    // NOTE(blt) I'm not sure if this is the right definition. I don't
+    // understand the purpose of container_size.
```
Manishearth: it's supposed to give you a size to use for a container. This would probably change with https://github.com/rust-fuzz/rust_arbitrary/issues/18

src/unstructured.rs
```diff
@@ -64,4 +65,56 @@ impl<'a> Unstructured<'a> {
...
+                inner: self.buffer[offset..].iter().cloned(),
+            })
+        }
+    }
+}
+
+/// An iterator of the remaining bytes returned by
+/// `Unstructured::take_rest`.
+pub struct TakeRest<'a> {
+    inner: iter::Cloned<slice::Iter<'a, u8>>,
```
Manishearth: use `iter::Copied` instead?

src/lib.rs
```diff
@@ -176,15 +176,19 @@ impl_arbitrary_for_floats! {
...
-        const SURROGATES_START: u32 = 0xD800;
-        let mut c = <u32 as Arbitrary>::arbitrary(u)? % CHAR_END;
+
+        // The largest u32 that is a valid unicode character.
+        const CHAR_END: u32 = 0x0011_000 - 1;
+
+        // The size of the surrogate blocks.
+        const SURROGATES_SIZE: u32 = 0xD800;
+
+        let c: u32 = u.int_in_range(0..=CHAR_END)?;
```
Manishearth: I'm not sure if we should do this, this seems slower.

Manishearth: Also, the older method slurps up chars as utf32, this will slurp them up as basically utf24 (not a thing that actually exists). The high byte in UTF32 is always null. This seems weird. If we want chars to parse as variable length we should attempt to parse them as utf8 instead.

src/unstructured.rs
```diff
@@ -44,11 +44,23 @@ impl<'a> Unstructured<'a> {
...
+            Some(0) => {
+                self.data = &[];
+                return Ok(0);
+            }
+            Some(max_size) => {
+                // Take lengths from the end of the data, since the `libFuzzer` folks
+                // found that this lets fuzzers more efficiently explore the input
+                // space.
+                //
+                // https://github.com/rust-fuzz/libfuzzer-sys/blob/0c450753/libfuzzer/utils/FuzzedDataProvider.h#L92-L97
```
Manishearth: this is cool

src/finite_buffer.rs
fitzgen: That was just code motion of an existing comment, FWIW.

src/unstructured.rs
fitzgen: I think I will actually remove `TakeRest` and just return the rest of the slice directly -- sound good?

src/lib.rs
fitzgen: Sounds good, will revert this commit.

Manishearth:
That works

PR #21, opened 2019-12-27
Add `Arbitrary::size_hint`

fitzgen:
r? @Manishearth 

Thanks for reviewing all my PRs :sweat_smile:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,6 +32,8 @@ pub mod unstructured;
 #[doc(inline)]
 pub use unstructured::Unstructured;
 
+pub mod size_hint;
+
 use std::borrow::{Cow, ToOwned};
 use std::cell::{Cell, RefCell, UnsafeCell};
 use std::collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque};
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -168,6 +170,30 @@ pub trait Arbitrary: Sized + 'static {
...
+    /// Get a size hint for how many bytes out of an `Unstructured` this type
+    /// needs to construct itself.
+    ///
+    /// This is useful for determining how many elements we should insert when
+    /// creating an arbitrary collection.
+    ///
+    /// The return value is similar to
+    /// [`Iterator::size_hint`][iterator-size-hint]: it returns a tuple where
+    /// the first element is a lower bound on the number of bytes required, and
+    /// the second element is an optional upper bound.
```
Manishearth: Should this be an optional upper bound, or an exact value? Will there be cases where having an upper bound is more useful than an exact value?

Manishearth: I don't think we support any types in this crate where that distinction matters so this just would involve changing the comment, but making it an exact value simplifies the next steps for the container stuff, I _think_

Manishearth: It actually might not matter at all. Your call.

Manishearth: Also I just read your comment on the issue. Never mind :)

PR #23, opened 2019-12-29
Add Arbitrary::arbitrary_take_rest

### LINKED ISSUE #17: Arbitrary impl for String isn't great
Even with https://github.com/rust-fuzz/rust_arbitrary/issues/8 fixed, we're not generating strings ideally. The current impl treats it as a vec of chars, which is _isomorphic_ to a string but not typically how a fuzzer will think about a string. libFuzzer will try to mutate inputs using strings and other patterns from the program, so it's important that on feeding the fuzz target `"banana"` in bytes, it sees the same string. Currently you'd need to pass the string `\0\0\0\0b\0\0\0\0a\0\0\0\0n...` along with a container size at the beginning (which we should ideally fix, see https://github.com/rust-fuzz/rust_arbitrary/issues/18)

We should do something more UTF8-aware (but, like the `char` impl, able to swiftly fall back)

### LINKED ISSUE #18: "Fill all available space" solution for container size

Manishearth:
Fixes #17, fixes #18

<s>This does not migrate arrays and tuples over, but it handles everything else.</s>

r? @fitzgen

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -9,6 +9,7 @@
 //! Wrappers around raw, unstructured bytes.
 
 use crate::{Arbitrary, Error, Result};
+use std::marker::PhantomData;
 use std::{mem, ops};
 
 /// A source of unstructured data.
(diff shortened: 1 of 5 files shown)


Manishearth:
One thing I noticed while writing tests for this: `arbitrary_len` consumes eight bytes. Since most fuzz strings are much shorter perhaps it should only consume as many bytes as necessary to index the fuzz string? E.g. for most short strings it only needs a `u8`, and _at best_ a `u16` for longer ones.

I'm not sure of this but also ideally if the parsed number fits in the range we should just use it without any of the bit shifting. The goal is that lengths should ideally be read directly off the end of the byte stream in the common case.

Manishearth:
Also, I just tested it, the fuzzer is now able to find string crashes _really_ quickly

src/unstructured.rs
```diff
@@ -390,6 +391,40 @@ impl<'a> Unstructured<'a> {
...
         mem::replace(&mut self.data, &[])
     }
+
+    /// Provide an iterator over elements for constructing a collection
+    ///
+    /// This is useful for implementing [`Arbitrary::arbitrary`] on collections
+    /// since the implementation is simply `u.arbitrary_iter()?.collect()`
+    pub fn arbitrary_iter<'b, ElementType: Arbitrary>(
+        &'b mut self,
+    ) -> Result<ArbitraryIter<'a, 'b, ElementType>> {
```
fitzgen: nice

src/unstructured.rs
```diff
@@ -406,6 +406,26 @@ impl<'a> Unstructured<'a> {
...
+    /// all the remaining bytes.
+    ///
+    /// This is useful for implementing [`Arbitrary::arbitrary_take_rest`] on collections
+    /// since the implementation is simply `u.arbitrary_take_rest_iter()?.collect()`
+    pub fn arbitrary_take_rest_iter<ElementType: Arbitrary>(
+        self,
+    ) -> Result<ArbitraryTakeRestIter<'a, ElementType>> {
+        let (lower, upper) = ElementType::size_hint();
+
+        let elem_size = upper.unwrap_or(lower * 2);
```
fitzgen: This needs a

```rust
let elem_size = std::cmp::max(1, elem_size);
```

to avoid div-by-zero

src/unstructured.rs
```diff
@@ -406,6 +406,26 @@ impl<'a> Unstructured<'a> {
...
+        let size = self.len() / elem_size;
+        println!("{:?}", size);
```
fitzgen: nitpick: remove this leftover debugging stuff

fitzgen:
> Also, I just tested it, the fuzzer is now able to find string crashes _really_ quickly

:tada:

Manishearth:
Now that https://github.com/rust-fuzz/arbitrary/pull/24 landed I added an additional test for collections.

PR #24, opened 2019-12-29
Improve Unstructured::arbitrary_byte_size to only pull as many bytes as needed

Manishearth:
See https://github.com/rust-fuzz/arbitrary/pull/23#issuecomment-569493197 . It feels unnecessary to pull eight whole bytes off of the buffer every time we need a length, when the length is usually never going to be more than a single byte. This changes `arbitrary_byte_size()` to only pull as many bytes as needed to represent all possible choices.

This also makes testing `Arbitrary` easier since there's a very predictable format to the byte string being parsed -- for smaller strings, the length is just the last byte (wrapped around if necessary)

r? @fitzgen

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -218,23 +218,46 @@ impl<'a> Unstructured<'a> {
     }
 
     fn arbitrary_byte_size(&mut self) -> Result<usize> {
-        match self.data.len().checked_sub(mem::size_of::<usize>()) {
-            None => return Err(Error::NotEnoughData),
-            Some(0) => {
-                self.data = &[];
-                return Ok(0);
-            }
-            Some(max_size) => {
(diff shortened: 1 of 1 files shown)


src/unstructured.rs
```diff
@@ -509,3 +532,17 @@ impl_int! {
...
+fn test_byte_size() {
+    let mut u = Unstructured::new(&[1, 2, 3, 4, 5, 6, 7, 8, 9, 6]);
+    // Should take one byte off the end
+    assert_eq!(u.arbitrary_byte_size().unwrap(), 6);
+    let mut v = vec![];
+    v.resize(260, 0);
+    v.push(1);
+    v.push(4);
+    let mut u = Unstructured::new(&v);
+    // Should read two bytes off the end
```
fitzgen: Do you want to assert these via the `len` method?

PR #26, opened 2020-01-13
Bump to version 0.3.0

### LINKED ISSUE #19: Tracking issue for 0.3.0 release

fitzgen:
Fixes #19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,77 @@
+## Unreleased
+
+Released YYYY-MM-DD.
+
+### Added
+
+* TODO (or remove section if none)
+
+### Changed
+
+* TODO (or remove section if none)
+
+### Deprecated
+
+* TODO (or remove section if none)
+
+### Removed
+
+* TODO (or remove section if none)
+
+### Fixed
+
+* TODO (or remove section if none)
+
+### Security
+
+* TODO (or remove section if none)
+
+--------------------------------------------------------------------------------
+
+## 0.3.0
+
+Released 2020-01-13.
+
+### Added
+
+* Added the `"derive"` cargo feature, to enable `#[derive(Arbitrary)]` for
+  custom types. Enabling this feature re-exports functionality from the
+  `derive_arbitrary` crate.
+* The custom derive for `Arbitrary` implements the shrink method for you now.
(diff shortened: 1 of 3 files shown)


fitzgen:
Gah I can't publish because I'm not an owner of `derive_arbitrary` on crates.io. @nagisa would you mind adding me as an owner? Thanks!

Manishearth:
I'm also okay with releasing this as arbitrary-derive instead, that's the typical naming scheme anyway. We can update the original with a redirect later

nagisa:
Sent out the invite.

fitzgen:
Thank you!

fitzgen:
Published!

PR #31, opened 2020-01-22
Add a `depth` parameter to `size_hint` to avoid infinite recursion

### LINKED ISSUE #30: SIGSEGV on certain values
I _think_ this is an issue with arbitrary, but let me know where I should move it if not.

I'm observing an issue where derived arbitrary types trigger a SIGSEGV when running cargo-fuzz. Here's a minimal example:

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;
use arbitrary::Arbitrary;

#[derive(Arbitrary, Clone, Debug, PartialEq)]
pub enum Op {
    A(A),
    B(B),
}

#[derive(Arbitrary, Clone, Debug, PartialEq)]
pub enum A {
    Leaf,
    B(Box<B>),
}

#[derive(Arbitrary, Clone, Debug, PartialEq)]
pub enum B {
    Leaf,
    A(Box<A>),
}

fuzz_target!(|ops: Vec<Op>| {
});
```

Running this on the latest nightly (both mac and linux) triggers a SIGSEGV:

```
INFO: Seed: 3703294504
INFO: Loaded 1 modules   (17415 inline 8-bit counters): 17415 [0x10f5ef708, 0x10f5f3b0f), 
INFO: Loaded 1 PC tables (17415 PCs): 17415 [0x10f5f3b10,0x10f637b80), 
INFO:       59 files found in /Users/yusuf/Desktop/sandbox/fuzz-bug/fuzz/corpus/fuzz_target_1
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
────────────────────────────────────────────────────────────────────────────────

Error: Fuzz target exited with signal: 11
```

Any of these changes will remove the segfault:

1) Changing the fuzz target's input from `Vec<Op>` to just `Op`.
2) Removing the `B` enum variant from `A`
3) Removing the `A` enum variant from `B`

fitzgen:
Note that the implementations of `size_hint` only need to increment the depth at indirection points where recursion can happen, e.g. at a `Box`. We take advantage of this in the impls for `std` types, but conservatively do not take advantage of that in the custom derive, since it doesn't have the necessary info.

Fixes #30

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,8 +202,62 @@ pub trait Arbitrary: Sized + 'static {
     /// default with a better implementation. The
     /// [`size_hint`][crate::size_hint] module will help with this task.
     ///
+    /// ## The `depth` Parameter
+    ///
+    /// If you 100% know that the type you are implementing `Arbitrary` for is
+    /// not a recursive type, or your implementation is not transitively calling
+    /// any other `size_hint` methods, you can ignore the `depth` parameter.
+    /// Note that if you are implementing `Arbitrary` for a generic type, you
+    /// cannot guarantee the lack of type recrusion!
+    ///
+    /// Otherwise, you need to use
+    /// [`arbitrary::size_hint::recursion_guard(depth)`][crate::size_hint::recursion_guard]
+    /// to prevent potential infinite recursion when calculating size hints for
+    /// potentially recursive types:
+    ///
+    /// ```
+    /// use arbitrary::{Arbitrary, Unstructured, size_hint};
+    ///
+    /// // This can potentially be a recursive type if `L` or `R` contain
+    /// // something like `Box<Option<MyEither<L, R>>>`!
+    /// enum MyEither<L, R> {
+    ///     Left(L),
+    ///     Right(R),
+    /// }
+    ///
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -702,8 +781,11 @@ impl<A: Arbitrary + Ord> Arbitrary for BinaryHeap<A> {
         u.arbitrary_take_rest_iter()?.collect()
     }
 
-    fn size_hint() -> (usize, Option<usize>) {
-        crate::size_hint::and(<usize as Arbitrary>::size_hint(), (0, None))
+    #[inline]
+    fn size_hint(depth: usize) -> (usize, Option<usize>) {
+        crate::size_hint::recursion_guard(depth, |depth| {
+            crate::size_hint::and(<usize as Arbitrary>::size_hint(depth), (0, None))
```
Manishearth: I don't see why we need it for the collections, we're not calling `A::size_hint` anywhere since we just lazily provide a minimum bound of `0` (plus `usize`)

fitzgen: good catch!

fitzgen:
@Manishearth did you want to take another look at this?

PR #37, opened 2020-03-18
Remove need for `synstructure` dependency

alexcrichton:
This dependency activates a number of features on `syn`, such as
`extra-traits`, which can cause excessive recompiles if `synstructure`
is part of some builds but not part of others. Especially for
`derive(Arbitrary)` it's often included in some builds (tests) but not
others (normal builds). By removing this dependency it should be a bit
speedier to build while also avoiding feature thrashing in normal builds.

### DIFF
--- a/derive/Cargo.toml
+++ b/derive/Cargo.toml
@@ -13,8 +13,8 @@ documentation = "https://docs.rs/arbitrary/"
 
 [dependencies]
 proc-macro2 = "1.0"
-syn = "1.0"
-synstructure = "0.12"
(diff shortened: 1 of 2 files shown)


fitzgen:
I'll publish a minor version bump

alexcrichton:
@fitzgen mind publishing a version of derive_arbitrary as well? I think the arbitrary crate may have been bumped but not this one?

fitzgen:
Woops! Fixed and published now.

PR #40, opened 2020-04-27
Use full path to `Arbitrary` trait in proc macro

### LINKED ISSUE #39: Derive failing in v0.4.2
# What is the issue

The derive macro seems to have build issues when called upon via the crate in version `0.4.2`

# Example 

Cargo.toml
```
[dependencies]
arbitrary = { version = "=0.4.2", features = ["derive"] }
```

main.rs
```rust
use arbitrary;

#[derive(arbitrary::Arbitrary)]
pub struct Wrapper {
    pub x: u8,
}

fn main() {
    let a = Wrapper {
        x: 3,
    };
    println!("{}", a.x);

```

# Version 0.4.1

This was working in v0.4.1 and can be seen by changing the Cargo.toml in the above example to 
```
arbitrary = { version = "=0.4.1", features = ["derive"] }
```
And re-running main.rs from above.

# libfuzzer-sys

I'm not sure the exact cause of the issue but it will impact `libfuzzer_sys v0.3.2` which 
re-exports `arbitrary::Arbitrary`

# Work around for `libfuzzer_sys v0.3.2`

I noticed there is a work around by doing 
Also have `arbitrary v0.4.2` as a dependency with `features = ["derive']`.
Then importing `Arbitrary` directly from `arbitrary`

e.g. for a fuzzer 

```
use libfuzzer_sys::fuzz_target;
use arbitrary::Arbitrary;
```

as opposed to how it would previously be done
```
use libfuzzer_sys::{fuzz_target, arbitrary::Arbitrary};
```

Manishearth:
Fixes #39

r? @fitzgen

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -130,7 +130,7 @@ fn gen_size_hint_method(input: &DeriveInput) -> TokenStream {
         let tys = fields.iter().map(|f| &f.ty);
         quote! {
             arbitrary::size_hint::and_all(&[
-                #( <#tys as Arbitrary>::size_hint(depth) ),*
+                #( <#tys as arbitrary::Arbitrary>::size_hint(depth) ),*
             ])
         }
     };


Manishearth:
@fitzgen can you publish a new version? I don't have access to publishing the derive

fitzgen:
Yep, I'll also add you on crates.io

fitzgen:
Done!

PR #48, opened 2020-06-18
Impl Arbitrary for RangeInclusive

c410-f3r:
Also relax bounds for both Range and RangeInclusive.

Is it possible to release a new version?


fitzgen:
0.4.5 published to crates.io!

PR #49, opened 2020-06-18
Impl Arbitrary for unbounded ranges

c410-f3r:
Thanks for releasing a new version. `RangeInclusive` was my use-case but now I realized that it is possible to generalize the implementation, so here it is.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -504,7 +504,13 @@ impl Arbitrary for AtomicUsize {
 }
 
 macro_rules! impl_range {
-    ($range:ty, $value_closure:expr, $value_ty:ty, $fun:ident($fun_closure:expr)) => {
+    (
+        $range:ty,
+        $value_closure:expr,
+        $value_ty:ty,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -503,64 +503,83 @@ impl Arbitrary for AtomicUsize {
...
-            <A as Arbitrary>::size_hint(depth),
-            <A as Arbitrary>::size_hint(depth),
-        )
-    }
+            #[inline]
+            fn size_hint(depth: usize) -> (usize, Option<usize>) {
+                crate::size_hint::and(
+                    <A as Arbitrary>::size_hint(depth),
+                    <A as Arbitrary>::size_hint(depth),
+                )
```
fitzgen, line 522: This should be just `<A as Arbitrary>::size_hint(depth)` for unbounded ranges, so I think this macro needs to be tweaked a little bit.

c410-f3r: Ooppsss...

PR #55, opened 2020-08-30
Add support for arbritrary arrays

c410-f3r:
`T: Default` would remove all `unsafe`s but it isn't currently possible for `[T; N]` and this bound limitation isn't desirable for stuff that don't implement `Default`.

If https://github.com/rust-lang/rust/pull/75644 is going to be accepted, then the auxiliary functions won't be necessary in the near future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -571,6 +571,8 @@ macro_rules! arbitrary_tuple {
 }
 arbitrary_tuple!(A B C D E F G H I J K L M N O P Q R S T U V W X Y Z);
 
+// Helper to safely create arrays since the standard library doesn't
+// provide one yet. Shouldn't be necessary in the future.
 struct ArrayGuard<T, const N: usize> {
     dst: *mut T,
     initialized: usize,
@@ -591,8 +593,10 @@ where
     F: FnMut(usize) -> T,
 {
     let mut array: mem::MaybeUninit<[T; N]> = mem::MaybeUninit::uninit();
+    let array_ptr = array.as_mut_ptr();
+    let dst = array_ptr as _;
     let mut guard: ArrayGuard<T, N> = ArrayGuard {
-        dst: array.as_mut_ptr() as _,
+        dst,
         initialized: 0,
     };
     unsafe {
@@ -610,8 +614,10 @@ where
     F: FnMut(usize) -> Result<T>,
 {
     let mut array: mem::MaybeUninit<[T; N]> = mem::MaybeUninit::uninit();
+    let array_ptr = array.as_mut_ptr();
+    let dst = array_ptr as _;
     let mut guard: ArrayGuard<T, N> = ArrayGuard {
-        dst: array.as_mut_ptr() as _,
+        dst,
         initialized: 0,
     };
     unsafe {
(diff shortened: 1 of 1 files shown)


fitzgen:
I would prefer not to have nightly-only features in this crate.

c410-f3r:
@fitzgen You might have a better insight but I am betting on 1.49  that is going to be released on 2020-12-31

fitzgen:
I don't have any insider insight here, but I'd prefer revisiting after it merges to stable. Happy to leave this PR open until then.

c410-f3r:
The time has come.

src/lib.rs
```diff
@@ -571,61 +571,85 @@ macro_rules! arbitrary_tuple {
...
-                    $( <$ts as Arbitrary>::size_hint(depth) ),*
-                ])
-            }
+fn create_array<F, T, const N: usize>(mut cb: F) -> [T; N]
+where
+    F: FnMut(usize) -> T,
+{
+    let mut array: mem::MaybeUninit<[T; N]> = mem::MaybeUninit::uninit();
+    let mut guard: ArrayGuard<T, N> = ArrayGuard {
+        dst: array.as_mut_ptr() as _,
```
fitzgen, line 595: What is the `as` cast doing here? Converting from `*mut [T; N]` to `*mut T`? If so, can we lift that to multiple `let`s that show the types, like this:

```rust
let dst: *mut [T; N] = array.as_mut_ptr();
let dst: *mut T = array as _;
// ...
```

src/lib.rs
```diff
@@ -571,61 +571,85 @@ macro_rules! arbitrary_tuple {
...
-                    Arbitrary::arbitrary(u)?,
-                    $(<$ts as Arbitrary>::arbitrary(u)?),*
-                ])
-            }
-
-            #[allow(unused_mut)]
-            fn arbitrary_take_rest(mut u: Unstructured<'a>) -> Result<[T; $n]> {
-                $(let $as = $ts::arbitrary(&mut u)?;)*
-                let last = Arbitrary::arbitrary_take_rest(u)?;
+struct ArrayGuard<T, const N: usize> {
```
fitzgen, line 574: Could you add a little documentation for what role this guard plays? That will help future readers of this code.

src/lib.rs
```diff
@@ -571,61 +571,85 @@ macro_rules! arbitrary_tuple {
...
+    }
+}
+
+fn try_create_array<F, T, const N: usize>(mut cb: F) -> Result<[T; N]>
+where
+    F: FnMut(usize) -> Result<T>,
+{
+    let mut array: mem::MaybeUninit<[T; N]> = mem::MaybeUninit::uninit();
+    let mut guard: ArrayGuard<T, N> = ArrayGuard {
+        dst: array.as_mut_ptr() as _,
```
fitzgen, line 614: Ditto regarding the `as` here.

fitzgen (commented):
Looks good to me, modulo a couple nitpicks below. Fix those up and then we can merge this! Thanks @c410-f3r!

c410-f3r:
Done. Thanks for the review

fitzgen:
Thanks!

PR #67, opened 2020-12-18
Implement Arbitrary for &[u8].

frewsxcv:
https://github.com/rust-fuzz/arbitrary/issues/62#issuecomment-736041629

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -626,6 +626,22 @@ arbitrary_array! { 32, (T, a) (T, b) (T, c) (T, d) (T, e) (T, f) (T, g) (T, h)
 (T, z) (T, aa) (T, ab) (T, ac) (T, ad) (T, ae) (T, af)
 (T, ag) }
 
+impl<'a> Arbitrary<'a> for &'a [u8] {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        let len = u.arbitrary_len::<u8>()?;
+        u.bytes(len)
+    }
+
+    fn arbitrary_take_rest(u: Unstructured<'a>) -> Result<Self> {
+        Ok(u.take_rest())
+    }
+
+    #[inline]
+    fn size_hint(depth: usize) -> (usize, Option<usize>) {
+        <usize as Arbitrary>::size_hint(depth)
+    }
+}
+
 impl<'a, A: Arbitrary<'a>> Arbitrary<'a> for Vec<A> {
     fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
         u.arbitrary_iter()?.collect()
@@ -1017,6 +1033,24 @@ mod test {
         let mut buf = Unstructured::new(&x);
         let expected = 1 | (2 << 8) | (3 << 16) | (4 << 24);
         let actual = i32::arbitrary(&mut buf).unwrap();
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    fn arbitrary_for_bytes() {
+        let x = [1, 2, 3, 4, 4];
+        let mut buf = Unstructured::new(&x);
+        let expected = &[1, 2, 3, 4];
+        let actual = <&[u8] as Arbitrary>::arbitrary(&mut buf).unwrap();
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    fn arbitrary_take_rest_for_bytes() {
+        let x = [1, 2, 3, 4];
+        let buf = Unstructured::new(&x);
+        let expected = &[1, 2, 3, 4];
+        let actual = <&[u8] as Arbitrary>::arbitrary_take_rest(buf).unwrap();
         assert_eq!(expected, actual);
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ Released YYYY-MM-DD.
 
 ### Added
 
-*
+* The `Arbitrary` trait is now implemented for `&[u8]`. [#67](https://github.com/rust-fuzz/arbitrary/pull/67)
 
 ### Changed
 
@@ -41,6 +41,8 @@ Released YYYY-MM-DD.
 ### Removed
 
 *
+
+--------------------------------------------------------------------------------
 
 ## 1.0.0-rc1
 


src/lib.rs
```diff
@@ -626,6 +626,17 @@ arbitrary_array! { 32, (T, a) (T, b) (T, c) (T, d) (T, e) (T, f) (T, g) (T, h)
...
 (T, ag) }
 
+impl<'a> Arbitrary<'a> for &'a [u8] {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        u.get_bytes(u.len())
+    }
+
+    #[inline]
+    fn size_hint(depth: usize) -> (usize, Option<usize>) {
+        crate::size_hint::and(<usize as Arbitrary>::size_hint(depth), (0, None))
```
frewsxcv, line 636: I cargo culted this line from the `Vec<A>` implementation. Can someone explain why we defer to the `usize` `Arbitrary` implementation? My initial idea was to just return `(0, None)` here, since we can construct a slice of bytes from nothing

nagisa: a4bbf9c1 added the method/impl to `Vec` cc @fitzgen

This impl effectively means `(8, None)` is always produced. Given the documentation on the `size_hint` method, I imagine `Arbitrary for Vec` would generate a `usize` length first, and then the number of elements or something along those lines. As thus the implementation needed at least 8 bytes of unstructured input to operate.

src/lib.rs
```diff
@@ -626,6 +626,17 @@ arbitrary_array! { 32, (T, a) (T, b) (T, c) (T, d) (T, e) (T, f) (T, g) (T, h)
 (T, z) (T, aa) (T, ab) (T, ac) (T, ad) (T, ae) (T, af)
 (T, ag) }
 
+impl<'a> Arbitrary<'a> for &'a [u8] {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        u.get_bytes(u.len())
```
nagisa, line 631: Consuming the entire unstructured buffer every time would mean that e.g.

```rust
#[derive(Arbitrary)]
struct Foo<'a> {
    bytes: &'a [u8],
    other_bytes: &'a [u9],
}
```

would always generate an empty slice for the 2nd field. You may want to do the `arbitrary_len` thing still here.

src/lib.rs
fitzgen: Yeah the `usize` size hint is for the vec's length.

src/lib.rs
fitzgen: Good eye @nagisa. Yes, we should still do the `arbitrary_len` thing, but also implement `arbitrary_take_rest` with this implementation that takes all the bytes.

src/lib.rs
frewsxcv: Thanks for clarifying! I think I understand the `usize` part now. And why do we use `size_hint::and(...)` here? Isn't this identical to just `<usize as Arbitrary>::size_hint(depth)`?

fitzgen: Yes it is. I'm not sure if I originally wrote this line or not, but in general I stylistically use `crate::size_hint::and(..)` even when its a no-op for nested fields or nested arbitrary calls just as a future proofing mechanism in case they grow in size sometime in the future or something. That doesn't really apply to this call, since it is using a hard coded value, rather than a nested `size_hint` call, so it should be safe to remove it.

frewsxcv: https://github.com/rust-fuzz/arbitrary/pull/67/commits/7d28b784114119f1b7a64f17e9c8361eb729905f

src/lib.rs
frewsxcv: > Consuming the entire unstructured buffer every time would mean that e.g.
> 
> ```rust
> #[derive(Arbitrary)]
> struct Foo<'a> {
>     bytes: &'a [u8],
>     other_bytes: &'a [u9],
> }
> ```
> 
> would always generate an empty slice for the 2nd field. You may want to do the `arbitrary_len` thing still here.

great catch! fixed in https://github.com/rust-fuzz/arbitrary/pull/67/commits/f951b5abbcafb8235670aa1afe426b535cdba81f

> Good eye @nagisa. Yes, we should still do the `arbitrary_len` thing, but also implement `arbitrary_take_rest` with this implementation that takes all the bytes.

the default implementation for `arbitrary_take_rest` defers to `arbitrary`:

https://github.com/rust-fuzz/arbitrary/blob/f951b5abbcafb8235670aa1afe426b535cdba81f/src/lib.rs#L173-L180

so seems like it'll already use this code. is that what you were thinking?

frewsxcv:
@nagisa @fitzgen Thanks for your patience, this is ready for another review. Just added a couple more commits

src/lib.rs
fitzgen: Not quite. It should be this:

```rust
impl<'a> Arbitrary<'a> for &'a [u8] {
    // ...
    fn arbitrary_take_rest(mut u: Unstructured<'a>) -> Result<Self> {
        Ok(u.take_rest())
    }
}
```

This will always consume the rest of the `Unstructured`'s input (which is what we want in the `_take_rest` method) without getting any kind of length prefix (which we want in regular `arbitrary`, but *not* in `arbitrary_take_rest`).

Does that clear things up? Sorry this required so many back and forths to spell out

src/lib.rs
```diff
@@ -1020,6 +1032,15 @@ mod test {
...
     }
 
+    #[test]
+    fn arbitrary_for_bytes() {
+        let x = [1, 2, 3, 4, 4];
+        let mut buf = Unstructured::new(&x);
+        let expected = &[1, 2, 3, 4];
+        let actual = <&[u8] as Arbitrary>::arbitrary(&mut buf).unwrap();
+        assert_eq!(expected, actual);
+    }
```
fitzgen, line 1042: Does this test still pass with the length prefix stuff? I expect it to pass when using `arbitrary_take_rest` but not with regular `arbitrary`, where I instead I would expect that we read a length (which is read from the end of the data) and then get some prefix of `[1, 2, 3]` depending on how `arbitrary_len` maps `4` into the `0..3` range.

fitzgen (commented):
LGTM modulo the `arbitrary_take_rest` stuff upthread and a question about the test below.

src/lib.rs
frewsxcv: Check the latest commit– does that match what's in your head?

fitzgen (approved):
Perfect! Thanks @frewsxcv :)

PR #70, opened 2021-01-24
Rename `Unstructured#get_bytes` to `Unstructured#bytes`

frewsxcv:
https://github.com/rust-fuzz/arbitrary/issues/62


fitzgen (approved):
Thanks!!

PR #79, opened 2021-03-25
Implement Arbitrary for nonzero integers

AljoschaMeyer:
Exactly what it says on the tin: `Arbitrary` implementations for the `NonZeroXY `types in [core::num](https://doc.rust-lang.org/std/num/index.html)


fitzgen (approved):
Thanks!

PR #80, opened 2021-05-17
Fix int_in_range using more bytes than necessary

e00E:
`offset` counts the number of bytes but the range is shifted as if it
counted the number of bits. This causes int_in_range to consume more
bytes than necessary.

The `int_in_range::<u32>(0..=u8::MAX)` case from the new test fails
without this commit.


fitzgen (approved):
Thanks!

PR #98, opened 2022-02-09
Add two new helper methods to `Unstructured`

fitzgen:
I've used these extensively in other test case generators. Should be generally helpful to anyone using this crate.

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -10,6 +10,7 @@
 
 use crate::{Arbitrary, Error, Result};
 use std::marker::PhantomData;
+use std::ops::ControlFlow;
 use std::{mem, ops};
 
 /// A source of unstructured data.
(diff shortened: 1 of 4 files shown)


src/unstructured.rs
```diff
@@ -523,6 +558,93 @@ impl<'a> Unstructured<'a> {
...
+    /// Call the given function an arbitrary number of times.
+    ///
+    /// The function is given this `Unstructured` so that it can continue to
+    /// generate arbitrary data and structures.
+    ///
+    /// You may optionaly specify minimum and maximum bounds on the number of
+    /// times the function is called.
+    ///
+    /// You may break out of the loop early by returning `Ok(false)`. You will
+    /// continue the loop by returning `Ok(true)`.
```
nagisa, line 571: There is a stable [std::ops::ControlFlow](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html) which seems much more appropriate here.

nagisa (approved):
Both of these helpers look nice and useful. +1 on landing ratio as is, but I feel loop can benefit from `ControlFlow` (at least in that it isn't necessary to document what the variants do…)

Manishearth (approved):

fitzgen:
Ah nice I didn't know about `ControlFlow`, let alone that it was stable. I'll switch to using that.

PR #101, opened 2022-02-14
Document size_hint() and fix implementations

### LINKED ISSUE #100: size_hint() has unclear semantics
The exact semantics of `size_hint()` are not clear from looking at the documentation and the existing impls. The lower bound specifically could have two different meanings:

- A hard limit. If there is less data available than specified, `arbitrary()` is expected to fail, don't even try to call it.
- A soft limit. All unique outputs can be achieved using larger inputs, so there is no advantage to using inputs shorter than the lower bound - though it may still work.

Additionally, it's unclear if the bounds apply only to `arbitrary()` or also to `arbitrary_take_rest()`. For more complex types, these implementations can differ significantly, where `arbitrary()` often needs some additional metadata bytes (increasing the lower bound), but `arbitrary_take_rest()` just passes the entire data on to its inner types.

Here are a few examples:

- [Integer types](https://github.com/rust-fuzz/arbitrary/blob/d0d238/src/lib.rs#L304) are `(size_of::<T>(), Some(size_of::<T>()))`. However, they are constructed using `Unstructured::fill_buffer()`, which if it "does not have enough underlying data to fill the whole buffer, it pads the buffer out with zeros.". Thus, it's possible to construct integers even from completely empty buffers, making this a soft limit.
- [`Vec<T>`](https://github.com/rust-fuzz/arbitrary/blob/d0d238/src/lib.rs#L691) is `(size_of::<usize>(), None)`. This is completely incorrect, but assuming `usize` should be `bool`, it does describe a soft limit for `arbitrary()`: the empty vector can be achieved both using the empty input as well as a single `false` continuation byte, so the empty input is redundant. **However**, for `arbitrary_take_rest()`, this does not hold: the only way to get an empty `Vec<u8>` is from the empty input - the size hint has to be `(0, None)`, which is also the hard limit for `arbitrary()`.
- [`&[u8]`](https://github.com/rust-fuzz/arbitrary/blob/d0d238d8/src/lib.rs#L676) is `(size_of::<usize>(), Some(size_of::<usize>()))`, which is even more wrong than `Vec`, but again `(size_of::<u8>(), None)` would be a correct soft limit for `arbitrary()` (the empty input is equivalent to a single length byte), but `(0, None)` is required for `arbitrary_take_rest()` or as an `arbitrary()` hard limit.

Xiretza:
This fixes #100.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,6 +202,16 @@ pub trait Arbitrary<'a>: Sized {
     /// default with a better implementation. The
     /// [`size_hint`][crate::size_hint] module will help with this task.
     ///
+    /// ## Invariant
+    ///
+    /// It must be possible to construct every possible output using only inputs
+    /// of lengths bounded by these parameters. This applies to both
+    /// [`Arbitrary::arbitrary`] and [`Arbitrary::arbitrary_take_rest`].
+    ///
+    /// This is trivially true for `(0, None)`. To restrict this further, it must be proven
+    /// that all inputs that are now excluded produced redundant outputs which are still
+    /// possible to produce using the reduced input space.
+    ///
     /// ## The `depth` Parameter
     ///
     /// If you 100% know that the type you are implementing `Arbitrary` for is
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1114,6 +1114,56 @@ impl<'a> Arbitrary<'a> for Ipv6Addr {
 mod test {
     use super::*;
 
+    /// Generates an arbitrary `T`, and checks that the result is consistent with the
+    /// `size_hint()` reported by `T`.
+    fn checked_arbitrary<'a, T: Arbitrary<'a>>(u: &mut Unstructured<'a>) -> Result<T> {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1114,6 +1114,53 @@ impl<'a> Arbitrary<'a> for Ipv6Addr {
...
+                "incorrect minimum size: indicated {}, actually consumed {}",
+                min,
+                consumed
+            );
+        }
+
+        result
+    }
+
+    fn checked_arbitrary_take_rest<'a, T: Arbitrary<'a>>(u: Unstructured<'a>) -> Result<T> {
```
fitzgen, line 1146: Let's add comments describing what these `checked_arbitrary*` test helpers are doing for future code readers.

fitzgen (commented):
Thanks! Just a little nitpick and then we can merge this.

src/lib.rs
Xiretza: Done!

fitzgen (approved):
Thanks!

PR #102, opened 2022-02-15
fix: Panic on 32-bit archs

ParkMyCar:
**Context**
Creating an [`arbitrary::Unstructured`](https://docs.rs/arbitrary/latest/arbitrary/struct.Unstructured.html) from a buffer larger than u16::MAX (~65Kb) will panic on 32-bit architectures (e.g. a Raspberry Pi).

The panic comes from choosing an arbitrary length for a collection. At one point we check if `self.data.len() <= std::u32::MAX as usize + 1`, and when `usize` is a `u32`, adding 1 causes an overflow and thus a panic. We do this check if our data is > 65Kb.

**Solution**
As a naive solution, this branch casts the necessary integers as `u64` to prevent the overflow. I confirmed locally that this does indeed fix the panic

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -236,19 +236,20 @@ impl<'a> Unstructured<'a> {
 
             // We only consume as many bytes as necessary to cover the entire
             // range of the byte string.
-            let len = if self.data.len() <= std::u8::MAX as usize + 1 {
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

PR #108, opened 2022-06-09
fix(Unstructured)!: don't produce meaningless data if exhausted

### LINKED ISSUE #107: Stack overflow for recursive type
The following code
```rust
use arbitrary::{Arbitrary, Unstructured};

#[derive(Debug, Arbitrary)]
enum Nat {
    Succ(Box<Nat>),
    Zero,
}

fn main() {
    Nat::arbitrary(&mut Unstructured::new(&[])).ok();
}
```
Results in a stack overflow

```None
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
fish: Job 1, 'cargo run' terminated by signal SIGABRT (Abort)
```
I guess there isn't any data for the discriminant to go on, so it choses the first one. Upon which it needs to construct a `Succ`, so it needs to construct a `Nat`. And there's the infinite recursion.

This issue is similar to #30.  
I have no practical use for this type, I just encountered it while helping someone on a Matrix channel. I know the issue can be side-stepped by reordering the variants.

Xiretza:
If the `Unstructured` is completely exhausted, returning a buffer completely filled with zeroes instead of indicating failure does not
make sense. It also causes issues for any users of `Arbitrary` implementations for integers (which are based on this method) who rely on eventually getting a nonzero result to e.g. break recursion.

Fixes #107.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1166,7 +1166,7 @@ mod test {
 
     #[test]
     fn finite_buffer_fill_buffer() {
-        let x = [1, 2, 3, 4];
+        let x = [1, 2, 3, 4, 5];
         let mut rb = Unstructured::new(&x);
         let mut z = [0; 2];
         rb.fill_buffer(&mut z).unwrap();
@@ -1174,7 +1174,9 @@ mod test {
         rb.fill_buffer(&mut z).unwrap();
         assert_eq!(z, [3, 4]);
         rb.fill_buffer(&mut z).unwrap();
-        assert_eq!(z, [0, 0]);
+        assert_eq!(z, [5, 0]);
+
+        assert!(rb.fill_buffer(&mut z).is_err());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


nagisa:
Do you have any measurements of how this affects fuzzing quality? Failing to produce any data sounds like we'd end up running a huge number of test runs that fail at producing a type out of input bytes.

Xiretza:
No, I don't have any data on this. Instinctively I would say that returning "fake" data instead of erroring may be faster in the short term, but would mean that the fuzzer takes a long time until it finds an input long enough to actually produce useful mutations.

This just seemed like the most obvious culprit for #107, though I supposed it could be worked around in other ways as well.

fitzgen:
We intentionally chose this behavior because empirically it helped `Arbitrary` implementations produce better test cases / increase fuzzing throughput.

See the commit message in https://github.com/rust-fuzz/arbitrary/commit/1d2f65360f634968b5525f4d6430a88a376277f0 for some more details.

Xiretza:
Do you have a better idea for a fix for #107 then?

jcaesar:
@fitzgen 
Should [this paragraph](https://github.com/rust-fuzz/arbitrary/blob/34be5b6d3d3bb513358e80453984652c88e07b75/src/lib.rs#L144-L148) be deleted from the docs, then?

fitzgen:
https://github.com/rust-fuzz/arbitrary/pull/113

PR #109, opened 2022-06-14
derive: Protect against unbounded recursion when `u.is_empty()`

fitzgen:
Fixes https://github.com/rust-fuzz/arbitrary/issues/107

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -12,7 +12,13 @@ pub fn derive_arbitrary(tokens: proc_macro::TokenStream) -> proc_macro::TokenStr
     let (lifetime_without_bounds, lifetime_with_bounds) =
         build_arbitrary_lifetime(input.generics.clone());
 
-    let arbitrary_method = gen_arbitrary_method(&input, lifetime_without_bounds.clone());
+    let recursive_count = syn::Ident::new(
+        &format!("RECURSIVE_COUNT_{}", input.ident),
+        Span::call_site(),
+    );
+
+    let arbitrary_method =
+        gen_arbitrary_method(&input, lifetime_without_bounds.clone(), &recursive_count);
     let size_hint_method = gen_size_hint_method(&input);
     let name = input.ident;
     // Add a bound `T: Arbitrary` to every type parameter T.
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

fitzgen:
I'll publish a new release with this fix.

fitzgen:
Published 1.1.1

derive/src/lib.rs
```diff
@@ -67,21 +77,51 @@ fn add_trait_bounds(mut generics: Generics, lifetime: LifetimeDef) -> Generics {
...
+) -> impl quote::ToTokens {
+    quote! {
+        #recursive_count.with(|count| {
+            if count.get() > 0 && u.is_empty() {
+                return Err(arbitrary::Error::NotEnoughData);
+            }
+
+            count.set(count.get() + 1);
+            let result = { #expr };
+            count.set(count.get() - 1);
```
jcaesar, line 92: Does this work as intended, i.e. is it ok to skip decreasing count if `#expr` aborts through `?`?

<details>
<summary>Generated code for my struct, for consideration</summary>

```rust
fn arbitrary(u: &mut Unstructured) -> Result<Self> {
    RECURSIVE_COUNT_Nah.with(|count| {
        if count.get() > 0 && u.is_empty() {
            return Err(arbitrary::Error::NotEnoughData);
        }
        count.set(count.get() + 1);
        let result = {
            Ok(
                match (u64::from(<u32 as Arbitrary>::arbitrary(u)?) * 2u64) >> 32 {
                    0u64 => Nah::Foo(
                        Arbitrary::arbitrary(u)?,
                        Arbitrary::arbitrary(u)?,
                        Arbitrary::arbitrary(u)?,
                    ),
                    1u64 => Nah::Bar(Arbitrary::arbitrary(u)?, Arbitrary::arbitrary(u)?),
                    _ => panic!("internal error: entered unreachable code"),
                },
            )
        };
        count.set(count.get() - 1);
        result
    })
}
```

</details>

jcaesar: Actually, thinking about it further, wouldn't it be better to only modify `count` when the unstructured is already empty?
* No need to access the thread local under normal conditions (performance micro-optimization, I know.)
* It's entirely possible that a recursive struct is benign and does finish generating on zeros only. But this currently aborts before trying that.

PR #114, opened 2022-07-03
Implement `Arbitrary` for `Rc<str>` and `Arc<str>`

Inky-developer:
These types did not implement it before, because `Arbitrary` is not implemented for `str`


Manishearth (approved):

Inky-developer:
Thanks, can you also approve the workflow run? Seems like I can't merge otherwise.

PR #128, opened 2022-10-09
Resolve clippy lints and add CI job

dtolnay:


Manishearth (approved):
not quite sure about the lazy eval lint but it doesn't matter too much

PR #160, opened 2023-10-30
Implement `Arbitrary` for `Arc<[A]>` and `Rc<[A]>`.

kpreid:
Also adjust the `Box<[A]>` implementation to be simpler and regular, not using a conversion from `Vec`, and provide an implementation of `arbitrary_take_rest()` rather than using the default.


fitzgen (approved):
Thanks!

fitzgen:
Published 1.3.2 with this change.

PR #171, opened 2024-01-17
Remove unused parameter in integer impl macro

krtab:
Hi!

`$unsigned` is not used anymore and can be removed.


fitzgen (approved):
Thanks!

PR #177, opened 2024-02-26
Add support for `no_std` compilation (updated)

pitdicker:
See discussion in #74.

This is useful to fuzz crates in their `no_std` configuration, which may utilize different code paths in such a crate.

I started by rebasing #74 but made a number of changes to split it in an `std` and `alloc` feature.


fitzgen:
This was the conclusion of #74: https://github.com/rust-fuzz/arbitrary/pull/74#issuecomment-902114435

Manishearth (approved):

nagisa:
@fitzgen counter question to the implementation constraints concern: is it really as bad as your comment makes it seem to restrict ourselves to libcore in the few circumstances where doing so would be necessary? I'm asking because `arbitrary` is largely a collection of `impl Arbitrary for ...`. Since most of the implementing types hail from `libstd`, those will be able to use std in their implementation. Outside of the `Arbitrary` trait and the few `impl`s for the basic libcore types, everything else (including `Unstructured`) can be made libstd-only and considered for libcore on a case-by-case basis. Wouldn't that be good enough?

src/lib.rs
```diff
@@ -14,6 +14,7 @@
 //! [`Arbitrary`](./trait.Arbitrary.html) trait's documentation for details on
 //! automatically deriving, implementing, and/or using the trait.
 
+#![cfg_attr(not(any(feature = "std", test)), no_std)]
```
fitzgen, line 17: To avoid dealing with different preludes that are vs aren't implicitly in scope depending on the features enabled, we should make this crate *always* be `no_std`. And then because the `std` feature implies the `alloc` feature, we can layer in imports like

```rust
#[cfg(feature = "alloc")]
use alloc::boxed::Box;
```

and we don't have to use any complex logic like `all(feature = "alloc", not(feature = "std"))` kinds of things.

My experience has also been that tools like `rust-analyzer` handle this pattern for `no_std` crates better as well.

fitzgen (commented):
I guess I won't be the one holding things back if everyone else wants this.

refcell:
Landing this would be highly appreciated.

Manishearth (approved):
in favor of landing once it's fixed

OKEAMAH (approved):

PR #178, opened 2024-03-15
impl Arbitrary for SystemTime

caspermeijn:


src/lib.rs
```diff
@@ -1272,6 +1273,34 @@ impl<'a> Arbitrary<'a> for SocketAddr {
     }
 }
 
+impl<'a> Arbitrary<'a> for SystemTime {
+    fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
+        // Create a SystemTime by adding or subtracting a duration from epoch.
+        // The result is not guaranteed to fit. Keep trying until a good SystemTime if found.
+        loop {
```
nagisa, line 1280: It seems unfortunate to have behaviour of this that is dependent on the target, but since we're going in that direction anyway, why not `#[cfg]` or even just check at runtime if the platform considers `UNIX_EPOCH` the minimal representable timestamp and just not bother with the add/sub on those targets?

caspermeijn: The docs for [`SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html) don't guarantee anything regarding what fits in a `SystemTime`. So I would not know what to check for. Maybe `UNIX_EPOCH` is the minimum. Maybe the year 0 is the minimum. 

Also, the maximum is interesting. It is not guaranteed that `SystemTime::UNIX_EPOCH + Duration::MAX` will fit. Therefore, it needs to be checked.

I could determine the minimum and maximum by trail and error, save this result and use that to make sure we never generate a value that doesn't fit. But that isn't great code as well.

nagisa: `UNIX_EPOCH.checked_sub(1).is_none()` implies that `UNIX_EPOCH` is the min value representable. This could be checked once and saved in a static or somesuch.

Unfortunately I've no good ideas for the maximum, so yeah, you would need the loop regardless. Feel free to ignore the comment until others chime in.

caspermeijn: As I understand, the minimum and maximum are platform-specific. I see three possible solutions:
1) Try to generate a value, check whether in range, if not, loop (current patch)
2) Try to find min and max, save those values, then generate a value in between (thus moving the loop to the process of finding the min and max)
3) Work with upstream to define the min and max as a constant (Duration has a min/max, so why shouldn't SystemTime)

PR #179, opened 2024-05-13
Add comment about output distribution

### LINKED ISSUE #68: Can Arbitrary be biased?
I have a question that might be clarified in the documentation. Is Arbitrary supposed to be exactly uniformly distributed? Or is it supposed to be quick? For example, I can generate a number between 0 and 10 by dividing the value of a `u8` by 26. With extra computations, the distribution could be made uniform, but is that necessary or desirable?

ThomasdenH:
This adds a note to the documentation of the `Arbitrary` trait to clarify the fact that uniformity is not required, but often desired. This closes #68.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -152,6 +152,17 @@ use std::sync::{Arc, Mutex};
 /// }
 /// # }
 /// ```
+///
+/// # A Note On Output Distributions
+///
+/// There is no requirement for a particular distribution of the values. For
+/// example, it is not required that every value appears with the same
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

fitzgen:
It looks like the trailing whitespace you introduced is causing CI to fail: https://github.com/rust-fuzz/arbitrary/actions/runs/9065479489/job/24914640202?pr=179#step:5:18

ThomasdenH:
I fixed it!

fitzgen:
Thanks!

PR #180, opened 2024-05-14
feature: Add f16 feature

FL33TW00D:
It would be very handy to be able to use f16 with Arbitrary.

We are integrating `f16` into `wgpu` which makes extensive use of this crate! https://github.com/gfx-rs/wgpu/pull/5701


fitzgen (commented):
I think this should go the other way: the `half` crate should add an optional `arbitrary` feature (the same way you wouldn't expect `serde` to add an optional `half` dependency, but `half` to add an optional `serde` dependency instead).

PR #181, opened 2024-05-31
Make usize's wire format independent of machine word size

Mrmaxmeier:
Hi,

I'm currently building Rust fuzzing targets for 32-bit as well as 64-bit platforms. (WASM, Native)
I'd like to be able to share the corpus/inputs between these builds but `arbitrary`'s wire format is not very portable at the moment. Full compatibility between inputs is not something we can or should strive to achieve, but this PR should cover all of the incompatibilities I've encountered. (Both direct use of `usize` in struct fields as well as indirect usage via `Vec<>` or similar containers)
Let me know if this is something you'd like to accept or reject.

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -359,13 +359,36 @@ impl_arbitrary_for_integers! {
     u32;
     u64;
     u128;
-    usize;
     i8;
     i16;
     i32;
     i64;
     i128;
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

PR #182, opened 2024-07-04
Implement choose_iter

matklad:
At times it is convenient to choose an item out of an iterator which isn't a slice!

Originally I wanted to use `Iterator`, rather than `ExactSizeIterator` bound here and implement simple reservoir sampling, but then reconsidered: reservoir sampling "wastes" entropy, and the amount of entropy we use is important for minimization. So I am going with ExactSize bound and leaving

```
fn choose_reservoir<T>(&mut self, choices: impl Iterator<Item=T>)
```

for someone's future work!

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -411,6 +411,39 @@ impl<'a> Unstructured<'a> {
         Ok(&choices[idx])
     }
 
+    /// Choose one of the given iterator choices.
+    ///
+    /// This should only be used inside of `Arbitrary` implementations.
+    ///
+    /// Returns an error if there is not enough underlying data to make a
+    /// choice or if no choices are provided.
+    ///
+    /// # Examples
+    ///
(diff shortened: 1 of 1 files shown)


matklad:
(my specific use case where I need this IRL (well, rather for [an argument over the internnet](https://lobste.rs/s/uutqvn/sad_state_property_based_testing)) is choosing random hash-map key)

matklad:
clippy failuer unrelated

nagisa:
Seems reasonable enough to me. If we were willing to release a breaking change, this could replace the slice-based `choose` as a strictly more general option. That's pretty unlikely due to `libc`-like reasons, however.

matklad:
I realized that it could have just one parameter and use I::Item in the return type. 

Not sure what’s better though-I kinda like redundant T, as it’s easier to specify to nudge type inference. So won’t change this unless you think one parameter would have been better

PR #185, opened 2024-07-23
Add faillible version of size_hint to properly handle recursive structures

### LINKED ISSUE #144: `size_hint` computation time blows up for "wide" recursive enums and structs
Calling `size_hint` on a "wide" recursive struct or enum is extremely slow. Examples:

```rust
#[derive(Debug, Arbitrary)]
enum Enum {
    None,
    A(Box<Enum>),
    B(Box<Enum>),
    C(Box<Enum>),
    D(Box<Enum>),
    E(Box<Enum>),
    F(Box<Enum>),
    G(Box<Enum>),
    H(Box<Enum>),
    I(Box<Enum>),
}


#[derive(Debug, Arbitrary)]
struct Struct {
    a: Box<Struct>,
    b: Box<Struct>,
    c: Box<Struct>,
    d: Box<Struct>,
    e: Box<Struct>,
    f: Box<Struct>,
    g: Box<Struct>,
    h: Box<Struct>,
    i: Box<Struct>,
}
```

This is because even though the `depth` parameter prevents this from going into an infinite loop, the derive implementation always calls `size_hint` once per field member even if they are of the same type. That makes the complexity close to $witdth^{recursion_depth}$, which here corresponds to $9^{20}$

Ideally the solution would be to somehow memoize the results of the calls to `size_hint` for each type but I don't see how to do that without having to add `Any` bounds everywhere. I tried implementing an unintrusive way to bail out when recursion is detected in https://github.com/sgued/arbitrary/tree/recursive-size-hint but it still doesn't solve every problem.

I think the most simple solution would be to make `size_hint` fallible and return an error in case of recursion. This would make it possible to bail out early for all implementations. It would still be easily possible to get the always correct `(0, None)` value by using `unwrap_or_default`. This would however be a breaking change.

Another solution I can think of would be to provide a way to override the `size_hint` implementation by something that always return `(0, None)`

sosthene-nitrokey:
The default implementation forwards to the current size_hint, so that it can be relied upon also for structures that still use the default implementation.

The default implementation of `size_hint` is not changed, so that it does not break existing implementations

Fix #144

### DIFF
--- a/src/foreign/core/cell.rs
+++ b/src/foreign/core/cell.rs
@@ -1,5 +1,5 @@
 use {
-    crate::{Arbitrary, Result, Unstructured},
+    crate::{Arbitrary, MaxRecursionReached, Result, Unstructured},
     core::cell::{Cell, RefCell, UnsafeCell},
 };
 
@@ -13,7 +13,12 @@ where
 
     #[inline]
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        <A as Arbitrary<'a>>::size_hint(depth)
+        Self::try_size_hint(depth).unwrap_or_default()
+    }
+
+    #[inline]
+    fn try_size_hint(depth: usize) -> Result<(usize, Option<usize>), MaxRecursionReached> {
+        <A as Arbitrary<'a>>::try_size_hint(depth)
     }
 }
 
@@ -27,7 +32,12 @@ where
 
     #[inline]
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        <A as Arbitrary<'a>>::size_hint(depth)
+        Self::try_size_hint(depth).unwrap_or_default()
+    }
+
+    #[inline]
+    fn try_size_hint(depth: usize) -> Result<(usize, Option<usize>), MaxRecursionReached> {
+        <A as Arbitrary<'a>>::try_size_hint(depth)
     }
 }
 
@@ -41,6 +51,11 @@ where
 
     #[inline]
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        <A as Arbitrary<'a>>::size_hint(depth)
+        Self::try_size_hint(depth).unwrap_or_default()
+    }
+
+    #[inline]
+    fn try_size_hint(depth: usize) -> Result<(usize, Option<usize>), MaxRecursionReached> {
+        <A as Arbitrary<'a>>::try_size_hint(depth)
     }
 }
--- a/src/foreign/core/num.rs
+++ b/src/foreign/core/num.rs
@@ -1,5 +1,5 @@
 use {
-    crate::{Arbitrary, Error, Result, Unstructured},
+    crate::{Arbitrary, Error, MaxRecursionReached, Result, Unstructured},
     core::{
         mem,
         num::{
@@ -131,6 +131,11 @@ where
 
     #[inline]
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        <A as Arbitrary<'a>>::size_hint(depth)
+        Self::try_size_hint(depth).unwrap_or_default()
+    }
+
+    #[inline]
+    fn try_size_hint(depth: usize) -> Result<(usize, Option<usize>), MaxRecursionReached> {
+        <A as Arbitrary<'a>>::try_size_hint(depth)
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,8 +39,9 @@ pub use derive_arbitrary::*;
 pub use unstructured::Unstructured;
 
 /// Error indicating that the maximum recursion depth has been reached while calculating [`Arbitrary::size_hint`]()
-#[derive(Debug, Clone, Copy)]
-pub struct MaxRecursionReached;
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct MaxRecursionReached {}
 
 impl core::fmt::Display for MaxRecursionReached {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,18 @@ pub use derive_arbitrary::*;
 #[doc(inline)]
 pub use unstructured::Unstructured;
 
+/// Error indicating that the maximum recursion depth has been reached while calculating [`Arbitrary::size_hint`]()
+#[derive(Debug, Clone, Copy)]
+pub struct MaxRecursionReached;
+
+impl core::fmt::Display for MaxRecursionReached {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str("Maximum recursion depth has been reached")
+    }
+}
+
+impl core::error::Error for MaxRecursionReached {}
+
 /// Generate arbitrary structured values from raw, unstructured data.
 ///
 /// The `Arbitrary` trait allows you to generate valid structured values, like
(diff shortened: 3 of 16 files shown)


fitzgen:
I'd have hoped that LLVM would inline the recursive `size_hint` calls up to the depth limit and effectively boiled it all away at compile time. Can you try adding `inline(always)` to the derived `size_hint` methods and seeing if that triggers the hoped-for behavior? I'd prefer not to complicate the trait interface anymore than it already is, if possible.

fitzgen:
My other thought is that, if we *do* need to define a new `size_hint` method, we don't need to make the whole thing fallible, just the potentially-recursive parts fallible. That means we can still provide a minimum bound (not a *maximal* minimal bound, but *a* minimal bound) even when we've reached the recursion depth. This would involve splitting the size hint into something like `my_size_hint` and `members_size_hint` (terrible names) where the former includes only the things that are guaranteed never to be recursive and the latter includes anything that might be recursive. Sort of like in a GC the difference between tracing your own shallow edges, vs recursively tracing the transitive closure of everything reachable from yourself, if that makes sense.

sosthene-nitrokey:
It's unlikely that LLVM would be able to inline a function call that is that slow, even though it could just be a constant calculation, 
and it would not happen in all optimization profiles.

Regarding keeping the lower bound: yes that is possible, by keeping it in the error type, but that would not be that useful, since this would still just return the first value that fails, and it would also complicate the implementation.

sosthene-nitrokey:
Rebased on top of main.

src/foreign/core/cell.rs
```diff
@@ -13,7 +13,12 @@ where
 
     #[inline]
     fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        <A as Arbitrary<'a>>::size_hint(depth)
+        Self::faillible_size_hint(depth).unwrap_or_default()
+    }
```
fitzgen, line 17: The old `size_hint` can be default-implemented with `fallible_size_hint` or the default right? No need to repeat this incantation in every implementation.

src/foreign/core/num.rs
```diff
@@ -12,7 +12,7 @@ use {
 macro_rules! impl_arbitrary_for_integers {
     ( $( $ty:ty; )* ) => {
         $(
-            impl<'a> Arbitrary<'a> for $ty {
+                impl<'a> Arbitrary<'a> for $ty {
```
fitzgen, line 15: This whitespace should be removed.

src/lib.rs
```diff
@@ -220,8 +232,73 @@ pub trait Arbitrary<'a>: Sized {
     /// create a better implementation. If you are writing an `Arbitrary`
     /// implementation by hand, and your type can be part of a dynamically sized
     /// collection (such as `Vec`), you are strongly encouraged to override this
-    /// default with a better implementation. The
-    /// [`size_hint`] module will help with this task.
+    /// default with a better implementation, and also override [`faillible_size_hint`]
+    ///
+    /// ## How to implement this
+    ///
+    /// If the size hint calculation is a trivial constant and does not recurse into any other `size_hint` call, you should implement it in `size_hint`:
```
fitzgen, line 239: Let's not muddy the waters by suggesting that sometimes people use the old size hint method and sometimes the new one.

Let's deprecate the old one and always recommend using the new one.

src/lib.rs
```diff
@@ -268,27 +385,35 @@ pub trait Arbitrary<'a>: Sized {
...
-    ///                 <R as Arbitrary>::size_hint(depth),
-    ///             )
+    ///         // `faillible_recursion_guard`.
+    ///         size_hint::faillible_recursion_guard(depth, |depth| {
+    ///            // If we aren't too deep, then `recursion_guard` calls
+    ///            // this closure, which implements the natural size hint.
+    ///            // Don't forget to use the new `depth` in all nested
+    ///            // `faillible_size_hint` calls! We recommend shadowing the
+    ///            // parameter, like what is done here, so that you can't
+    ///            // accidentally use the wrong depth.
```
fitzgen, line 403: This changed to use a 3-space indent, but should continue to use a 4-space indent.

src/lib.rs
```diff
@@ -268,27 +385,35 @@ pub trait Arbitrary<'a>: Sized {
...
     ///     }
     /// }
     /// ```
+    /// [iterator-size-hint]: https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.size_hint
+    /// [`size_hint`]: Arbitrary::size_hint
     #[inline]
-    fn size_hint(depth: usize) -> (usize, Option<usize>) {
-        let _ = depth;
-        (0, None)
+    fn faillible_size_hint(depth: usize) -> Result<(usize, Option<usize>), MaxRecursionReached> {
```
fitzgen, line 415: Nitpick: "faillible" is a French word, the English is "fallible". So we should rename this method to `fallible_size_hint` or perhaps `try_size_hint`.

fitzgen (commented):
LGTM modulo some inline comments that should be addressed before this can land. Thanks!

src/foreign/core/cell.rs
sosthene-nitrokey: Not really,  because of backwards compatibility.

If we make the default `size_hint` delegate to `faillible_size_hint`, this means we can't provide a default implementation for `faillible_size_hint` that calls back to `size_hint`, otherwise there would be an infinite loop.

We *could* have a default `faillible_size_hint` implementation that returns `Ok((0, None))`,  but then, when libraries migrate to calling `faillible_size_hint`, it will break for out of date dependencies that only implement `size_hint`.

src/lib.rs
sosthene-nitrokey: See https://github.com/rust-fuzz/arbitrary/pull/185#discussion_r1759119669

We could say that this is "just" an optimization and that it's ok to break it with a minor release.
Arguably a lot of users use the derive macro or don't implement size_hint, so it could be justified. And marking it as deprecated would make this pretty visible to downstream.

I also don't like the duplicate API.

src/lib.rs
sosthene-nitrokey: `try_size_hint` is better indeed.

src/lib.rs
sosthene-nitrokey: Good eye!

src/lib.rs
```diff
@@ -38,6 +38,18 @@ pub use derive_arbitrary::*;
 #[doc(inline)]
 pub use unstructured::Unstructured;
 
+/// Error indicating that the maximum recursion depth has been reached while calculating [`Arbitrary::size_hint`]()
+#[derive(Debug, Clone, Copy)]
+pub struct MaxRecursionReached;
```
fitzgen, line 43: For future-proofing, let's make this have a private field and avoid deriving `Copy`:

```suggestion
#[derive(Debug, Clone)]
pub struct MaxRecursionReached {
    _private: ()
}
```

src/lib.rs
fitzgen: The other comment about defaults and backwards compat makes sense, thanks for pointing that out.

That said, we should suggest that everyone move over to `try_size_hint` and ignore `size_hint` and/or implement it with `try_size_hint().unwrap_or_default()` and then avoid mentioning the deprecated method again.

src/lib.rs
sosthene-nitrokey: Why use a `_private` field instead of `#[non_exhaustive]` ?

sosthene-nitrokey: Should we also implement `Default` for it ?

src/lib.rs
sosthene-nitrokey: One of the things with my approach is that it does not require dependents to change what they call.

For example `libfuzzer_sys` does not care about the size hint being fallible, so it would just have been able `size_hint` and let the implementation deal with recursion internally.

I think the ideal setup would be to only have `try_size_hint` be part of the trait, and have an infallible `size_hint` in an extension trait (so users are not given the opportunity to override it) that does not even take a depth parameter for use in `libfuzzer_sys` for example. But that seems like adding more complexity and bring in naming conflicts.

sosthene-nitrokey: The backward compatibility issues will also mean that users will need to implement `size_hint` to call `try_size_hint` instead, and have to silence the `deprecated` warning to do so.

I think the issue this fixes most likely only affects users of the `derive` macros, and therefore I wanted to minimize the impact on other users. Especially given that a lot of types that can implement `arbitrary` will likely implement it on single, non recursive types and therefore they could still implement only `size_hint`.

Also, it's still possible to add deprecation attributes in the future if this does not happen to be the case.

src/lib.rs
fitzgen: > Why use a `_private` field instead of `#[non_exhaustive]` ?

No reason, I just always forget new Rust features.

fitzgen: > Should we also implement `Default` for it ?

If there isn't a strong reason to implement it, then I'd prefer to avoid it. It gives us more freedom down the line if we decide to do something like add the recursion depth or type names or something to the error, in which case a default might be tricky to choose.

sosthene-nitrokey: Without default, the only way for users to create it is to use `size_hint::try_recursion_guard`. Maybe there's some weird use case where someone wants to return this as an error without going through this function.

fitzgen: I think that's probably okay, if someone needs it then we can cross that bridge at that time.

fitzgen (approved):
Thanks!

Issue #44, opened 2020-05-16
Reversed arbitrary


epilys:
Hello,

I've been toying with custom mutators in cargo-fuzz with libfuzzer and it seems to me to support changes to typed data instead of bytes, the reversed arbrirtary operation would be really helpful. Here's the interface I have implemented so far:

```rust
/// Define a custom fuzz mutator.
///
/// If `$bytes` exceeds `$max_size`, it will be silently truncated.
///
/// ## Example
/// ```no_run
/// #![no_main]
/// use libfuzzer_sys::{fuzz_target, fuzz_mutator, llvm_fuzzer_mutate};
///
/// fuzz_target!(|data: &[u8]| {
///     let _ = std::str::from_utf8(data);
/// });
///
/// fuzz_mutator!(|data: &mut [u8], max_size: usize| {
///     println!("custom mutator called with data len = {} and max_size = {}", data.len(), max_size);
///     /* call wrapper function of libfuzzer's default mutator */
///     llvm_fuzzer_mutate(data, max_size)
/// });
/// ``` 
```

The mutator with typed data would be the equivalent:
```rust
fuzz_mutator!(|data: &mut T, max_size: usize | {
   loop {
   /* perform changes on `data` */

   if data.arbitrary_size() <= max_size { /* where arbitrary_size is some method from the Trait to calculate size in bytes */
     break;
  }
}
 });
```

Does this sound like a reasonable approach to you?

fitzgen:
So the idea is that this would also involve adding something like a `as_arbitrary_bytes` method on the `Arbitrary` trait? And the `fuzz_mutator!` would construct the `T: Arbitrary` for you, let you mutate it, and then call `as_arbitrary_bytes` to give the bytes back to libfuzzer?

And we would want that `x == T::arbitary_take_rest(x.as_arbitrary_bytes())` where `x: T`? (semi-aside: it might be hard to maintain this property for our arbitrary-length-getting functions)

This seems like a nice thing to have, but I haven't totally thought through how it might play out in terms of interaction with the `Arbitrary` trait, how nice we can keep the ux, and how well things interact and compose in practice.

However, a first step that I feel is safe to make without answering all those unknowns is to add a `libfuzzer_sys::fuzz_mutator!(|data: &mut  [u8], max_len: usize| { ... })` macro, that just supports `[u8]` and does not support `T: Arbitrary`.

fitzgen:
[...]

PR #187, opened 2024-08-04
Dearbitrary

LeoDog896:
Implements #44.

I would greatly appreciate feedback on naming the videos.

- `dearbitrary` intentionally builds `Unstructured` in reverse to counter the problems encountered in #94 (the slice `length` optimizations) by ensuring that size is built in reverse.
- `kani` is used to fastly assure primitive dearbitary/rearbitrary cases. I am still proof-checking the other primitives and will write fuzzing tests for the non-verifiable implementations.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,13 @@ jobs:
       run: cargo test --verbose --all-features --all
     - name: Build Examples
       run: cargo build --examples --all-features --all
+  kani:
+    name: Kani Verification
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - run: rustup update
+    - uses: model-checking/kani-github-action
   clippy:
     name: Clippy
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,6 @@ members = ["./fuzz"]
 
 [dev-dependencies]
 exhaustigen = "0.1.0"
+
+[target.'cfg(kani)'.dependencies]
+paste = "1.0"
--- a/src/back_note.md
+++ b/src/back_note.md
@@ -1,0 +1,2 @@
+Length information is generally encoded here.
+See [UnstructuredBuilder::extend_from_dearbitrary_iter_rev_with_length]
--- a/src/dearbitrary/error.rs
+++ b/src/dearbitrary/error.rs
@@ -1,0 +1,41 @@
+use std::{error, fmt};
+
+/// An enumeration of dearbitrartion errors
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Error {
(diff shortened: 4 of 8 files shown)


fitzgen:
Apologies for how long its taken me to respond to this PR.

I'm concerned about the complexity that this feature brings. I'm not convinced that it is worth it, just to be able to seed a corpus.

If seeding a corpus is an important use case, it can be done today by changing how the structure-aware fuzzing is done from a generative paradigm to a mutation paradigm:

* take bytes as the input to the fuzz target

* try to deserialize the bytes into your input type with `bincode::deserialize::<MyInputType>(fuzzer_input)` or similar (protobuf, etc...)

  * if the deserialization fails, continue to the next test case

  * if it succeeds, then run the structured input through your oracles, same as you otherwise would with the `arbitrary` crate

* additionally define a `fuzz_mutator!` that does something like

  ```rust
  let mut input = bincode::deserialize::<MyInputType>(input_bytes)
      .unwrap_or_else(|| MyInputType::default());

  // Mutate a test case. Not any more difficult to implement
  // than a by-hand `Arbitrary` implementation.
  input.mutate(seed);

  return bincode::serialize(input);
  ```

* now you can seed your corpus by manually building a bunch of `MyInputType`s and then `bincode::serialize` them into the corpus directory

(But backing up: It should also be noted that randomly generating a corpus before you start fuzzing isn't going to be expected to do any better than incrementally building the corpus from nothing. Seeding the corpus is only useful if you already have inputs that you know are interesting for other reasons, for example they've triggered bugs/crashes before or are Real World snippets of your programming language or etc...)

fitzgen:
In general, the mutation paradigm has another benefit over the generative paradigm as well: if you add knobs to only do shrinking/simplifying mutations, then it is trivial to build a test case minimizer on top of that which performs better than `cargo fuzz tmin`. And for things like programming languages, if you add a semantics-preserving mutation mode, then you can do differential fuzzing where you check the correctness of your various compiler passes and optimizations like

```rust
let input = the_fuzzer_input();
let result = run(&input);

let alt_input = input.mutate_preserving_semantics();
let alt_result = run(&alt_input);

assert_eq!(
    result,
    alt_result,
    "running two semantically-equivalent programs should produce the same results",
);
```

fitzgen:
(we should probably have a crate that is the mutation-paradigm-equivalent of the `arbitrary` crate in the `rust-fuzz` org)

LeoDog896:
By the way, my use of seeding was incorrect. I meant more for large arbitrary-derived structures, as developing an initial corpus is painstaking. Still, without it, the initial fuzzing warmup takes longer. However, the inconvenience also means that some libraries choose to go without a corpus for those types of structures.

While I do have a limited understanding of fuzzers, aren't most fuzzers mutative instead of generative already? Wouldn't methods like `mutate` or `mutate_preserving_semantics` have to explore much more slowly given that their inputs are dependent on the structure of `input`?

PR #188, opened 2024-08-12
feat(derive): add variant-attribute `#[arbitrary(skip)]`

sivizius:
This commit introduces the first variant-attribute: `skip`. Its purpose is to skip variants, that should not be generated. E.g. they might be required for parsing objects, but should not be serialized (e.g. `#[serde(skip_serializing)]`).

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -6,11 +6,14 @@ use syn::*;
 
 mod container_attributes;
 mod field_attributes;
+mod variant_attributes;
+
 use container_attributes::ContainerAttributes;
 use field_attributes::{determine_field_constructor, FieldConstructor};
+use variant_attributes::not_skipped;
 
-static ARBITRARY_ATTRIBUTE_NAME: &str = "arbitrary";
-static ARBITRARY_LIFETIME_NAME: &str = "'arbitrary";
+const ARBITRARY_ATTRIBUTE_NAME: &str = "arbitrary";
+const ARBITRARY_LIFETIME_NAME: &str = "'arbitrary";
 
 #[proc_macro_derive(Arbitrary, attributes(arbitrary))]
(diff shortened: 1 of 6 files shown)


fitzgen (commented):
Thanks!

Can you add a test like

```rust
enum Foo {
    A,
    #[arbitrary(skip)]
    B
}

for b in 0..=u8::MAX {
    let u = Unstructured::new(&[b]);
    let foo = Foo::arbitrary_take_rest(u).unwrap();
    assert!(!matches!(foo, Foo::B));
}
```

What about when there is only one variant and it is marked `skip`?

sivizius:
Empty `enum`s, those without any variant, that is not marked `skip`, including those without any variant at all, will now result in a compiler error, because there is no way to construct a valid value using this crate. This however, is somewhat beyond the initial scope of this PR.

derive/src/lib.rs
```diff
@@ -6,11 +6,14 @@ use syn::*;
...
+mod variant_attributes;
+
 use container_attributes::ContainerAttributes;
 use field_attributes::{determine_field_constructor, FieldConstructor};
+use variant_attributes::not_skipped;
 
-static ARBITRARY_ATTRIBUTE_NAME: &str = "arbitrary";
-static ARBITRARY_LIFETIME_NAME: &str = "'arbitrary";
+const ARBITRARY_ATTRIBUTE_NAME: &str = "arbitrary";
+const ARBITRARY_LIFETIME_NAME: &str = "'arbitrary";
```
sivizius, line 16: Technically unrelated, but no need to use `static` here.

fitzgen (approved):
Thanks!

sivizius:
There are remaining clippy-errors, which are outside of my changes. Should I fix them as well or can it be merged without?

sivizius:
I opened #190, perhaps it should be merged (assuming this fixes these errors) first?

fitzgen:
Thanks again!

PR #189, opened 2024-08-22
refactor(foreign): move implementations to dedicated, structured directory

sivizius:
At the moment, the trait-definition and all implementations for foreign types are located in `lib.rs` with 1732 LOC. IMHO this is difficult to maintain, especially when adding additional implementations. By moving all implementations into the private submodule `foreign`, sorting them by `core`, `alloc` and `std` and submodules similar to those documented in e.g. https://doc.rust-lang.org/std/, I see i.a. the following advantages:

1. Easier to find for which types the trait is implemented, which implementations are missing,
2. If one wants to add a implementation, one can just place it in the module according to its type definition in `core`/`alloc`/`std` and does not have to guess, whether to place it at the beginning or end of `lib.rs` or somewhere random,
3. It will be easier to feature-gate some implementations, e.g. `std` for `std`-implementations, which should be a default-feature, which can be disabled for `no_std`-builds,
4. Implementations for additional (common) crates, e.g. `chrono` or `time`, can be placed in `foreign` behind feature-gates,
5. Changes to one file does not affect changes to other files, so merging multiple PRs might be less conflicting

However, this is indeed quite a large change and will require additional PRs e.g. for feature-gating, fixing the usage of `std` in `alloc`, adding missing implementations, ….


fitzgen (approved):
Thanks!

PR #190, opened 2024-08-22
fix(unstructured): fix clippy-errors

sivizius:
See https://github.com/rust-fuzz/arbitrary/actions/runs/10476910752/job/29017347246?pr=188


fitzgen (approved):
Thanks!

PR #192, opened 2024-08-23
[WIP] feat: introduce ArbitraryInRange

sivizius:
This PR introduces the trait `ArbitraryInRange`, which allows limiting the range of possible values for some types. The intention is to use this trait in an not-yet-implemented `[arbitrary(range(min = …, max = …))]` field-attribute. A separate trait is necessary, because 1. not all types can be meaningful used as ranges, 2. I like to use `#[arbitrary(range(min = 23, max = 42))]` on `Option<u8>`. 

However, there are certainly some implementations missing, so still WIP.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -11,6 +11,8 @@ pub enum Error {
     NotEnoughData,
     /// The input bytes were not of the right format
     IncorrectFormat,
+    /// Cannot generate value in given range
+    InvalidRange,
 }
 
 impl fmt::Display for Error {
@@ -28,6 +30,7 @@ impl fmt::Display for Error {
                 f,
                 "The raw data is not of the correct format to construct this type"
             ),
+            Error::InvalidRange => write!(f, "Cannot generate a value in given range"),
         }
     }
 }
--- a/src/foreign/alloc/boxed.rs
+++ b/src/foreign/alloc/boxed.rs
@@ -1,39 +1,10 @@
 use {
-    crate::{size_hint, Arbitrary, Result, Unstructured},
+    crate::{Arbitrary, Result, Unstructured},
     std::boxed::Box,
 };
 
-impl<'a, A> Arbitrary<'a> for Box<A>
-where
-    A: Arbitrary<'a>,
-{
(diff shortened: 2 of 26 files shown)


yancyribbens:
I'd like to see this get added as well.  I've tested `impl<'a> ArbitraryInRange<'a>` and it worked.

yancyribbens:
Actually, the trait `ArbitraryInRange` doesn't appear to be working as intended.  For example, I would expect the following test to pass?

```rust
#[test]                                                                                                                                                                    │On branch add-test-to-in-range
fn arbitrary_in_range_for_integers() {                                                                                                                                     │Changes not staged for commit:
    let x = [1, 2, 3, 4];                                                                                                                                                  │  (use "git add <file>..." to update what will be committed)
    let mut u = Unstructured::new(&x);                                                                                                                                     │  (use "git restore <file>..." to discard changes in working directory)
    let result_one = i32::arbitrary_in_range(&mut u, &(0..=10));                                                                                                           │        modified:   src/tests.rs
                                                                                                                                                                           │
    let y = [1, 2, 3];                                                                                                                                                     │Untracked files:
    let mut u = Unstructured::new(&y);                                                                                                                                     │  (use "git add <file>..." to include in what will be committed)
    let result_two = i32::arbitrary_in_range(&mut u, &(0..=10));                                                                                                           │        src/.tests.rs.swp
                                                                                                                                                                           │        tags
    assert_ne!(result_one, result_two);                                                                                                                                    │
}
```

The result is always `0` (the first value in the range) no matter what u is given.

sivizius:
> The result is always 0 (the first value in the range) no matter what u is given.

In `src/foreign/core/num.rs` in macro `impl_arbitrary_for_integers`, I wrote:

```rust
let minimum = map_bound(range.start_bound(), |x| *x as Self);
let maximum = map_bound(range.start_bound(), |x| *x as Self);
u.int_in_range((minimum, maximum))
```

Note the second `start_bound` and the absence of `end_bound` m)

OKEAMAH (approved):

PR #193, opened 2024-08-31
Add Arbitrary in range to trait

yancyribbens:
Looking for the ability to create an Arbitrary type from a range.

For example, Given a Type:

```rust
pub struct A(u64);
```

Then a type can be created from range with:
```rust
impl<'a> Arbitrary<'a> for A {
    fn arbitrary_in_range<T>(u: &mut Unstructured<'a>, range: ops::RangeInclusive<T>) -> Result<Self>
    where
        T: arbitrary::unstructured::Int,
    {
        let a = u.int_in_range(range)?.to_u64();
        Ok(A(a))
    }
}
```

Marking this as draft to request feedback about adding.  Cheers.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,7 @@ pub mod size_hint;
 use core::array;
 use core::cell::{Cell, RefCell, UnsafeCell};
 use core::iter;
-use core::mem;
+use std::{mem, ops};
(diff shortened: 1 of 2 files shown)


yancyribbens:
I just noticed https://github.com/rust-fuzz/arbitrary/pull/192 which overlaps with this PR.

Issue #139, opened 2023-01-17
Defaulting to all-zeroes is unexpected


SimonSapin:
When `Unexpected` runs out of input data, `int_in_range` always returns `range.start()` (in my case, zero). This appears to be deliberate (https://github.com/rust-fuzz/arbitrary/commit/38192741b9a1f14424106545f4b3b1e56fb0a3d3, https://github.com/rust-fuzz/arbitrary/commit/1d2f65360f634968b5525f4d6430a88a376277f0) but was very unexpected to me, especially when `arbitrary::Error::NotEnoughData` exists.

This turned out to be involved in an infinite loop bug: https://github.com/apollographql/apollo-rs/issues/426

(FWIW I ran into this quickly because I’m using input bytes not from a fuzzer but from a fixed-seed PRNG in order to generate a deterministic benchmarking corpus. Because I don’t have a useful `size_hint`, I picked an… arbitrary… number of input bytes.)

Silently devolving into pretty much `Default` seems very bad when the point of this crate is to generate varied values. Would you consider reverting the decision not to exit early?

If not, is `Unstructured::is_empty` a good indicator that defaulting happened? (I could maybe try doubling the input size until `is_empty()` stays false after generating items)

fitzgen:
We've had empirical results that suggest that this behavior results in better fuzzing performance and exploration of the input state space, so I don't really want to revisit the decision.

> is `Unstructured::is_empty` a good indicator that defaulting happened?

Yes, you can use this method to determine when you're generators have consumed all the available entropy and will start to / have started to receive default values.

SimonSapin:
Instead of padding with zeros, would it make sense to cycle through already-used entropy?

fitzgen:
This was something that I considered in the past, but I think it makes mutation/reduction more difficult, since now mutating one byte doesn't make one change in the generated structure. For example, if you're trying to create a minimal reproducer, the one byte that causes the generator to generate the thing that triggers the bug could be tied to also generating another thing, and you'll never be able to reduce to the actual minimal reproducer.

`wasm-opt -ttf` (a Wasm test case generator that is part of the binaryen project) does this recycling of the input, and I've found that its made test case reduction very difficult when I've tried to reduce its inputs.

mkforsb:
[...]

PR #195, opened 2024-09-10
Add `int_in_range_checked`

mkforsb:
Adds a function `Unstructured::int_in_range_checked` that returns `Error::NotEnoughData` if there is not enough underlying data.

See #139.

### DIFF
--- a/src/unstructured.rs
+++ b/src/unstructured.rs
@@ -273,6 +273,10 @@ impl<'a> Unstructured<'a> {
     /// Do not use this to generate the size of a collection. Use
     /// `arbitrary_len` instead.
     ///
+    /// Do not use this function if you expect an error when there
+    /// is not enough underlying data to produce the requested type.
+    /// Use `int_in_range_checked` instead.
+    ///
     /// # Panics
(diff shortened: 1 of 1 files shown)


fitzgen (commented):
Can you clarify some more about the intent here? Because basically all functions will default to returning `0` (or the context-specific equivalent) when the underlying data has been exhausted, and this has been empirically found to produce better fuzzing throughput/coverage, and I fear that this is a "give a mouse a cookie" situation where if we add this variant then every single method in the library will eventually need to have a "checked" variant. You can also do this check downstream, without changes to the crate, today by checking `u.len()` and/or `u.is_empty()` (the latter would allow you to consume partial entropy, if there isn't quite enough for a full integer). So, altogether, I am pretty hesitant about this addition.

mkforsb:
The intent behind this particular PR was simply to provide the things I suggested in #139 rather than just leave it as a suggestion. I see what you mean about the other functions and I understand not wanting to add "checked" variants for every one. I'll solve my use case with an extension trait.

PR #196, opened 2024-09-10
docs: Use automatic link resolution of rustdoc

caspermeijn:
`rustdoc` will automatically resolve links to types that are in scope. Remove explicit link target when the type is already in scope.

Solves these warnings when running `cargo doc`:
```
warning: redundant explicit link target
   --> src/lib.rs:202:26
    |
202 |     /// [`Unstructured`][crate::Unstructured].
    |          --------------  ^^^^^^^^^^^^^^^^^^^ explicit target is redundant
    |          |
    |          because label contains path that resolves to same destination
    |
    = note: when a link's destination is not specified,
            the label is used to resolve intra-doc links
help: remove explicit link target
    |
202 |     /// [`Unstructured`].
    |         ~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,10 +8,10 @@
 
 //! The `Arbitrary` trait crate.
 //!
-//! This trait provides an [`Arbitrary`](./trait.Arbitrary.html) trait to
+//! This trait provides an [`Arbitrary`] trait to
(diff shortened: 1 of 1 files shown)


fitzgen (approved):
Thanks!

PR #210, opened 2024-12-05
Implement Arbitrary for core::cmp::Reverse

swwu:
We have a lot of data structures that depend on ordering requirements that we enforce via things like `BinaryHeap<Reverse<K>, V>`, so having `Reverse<T>: Arbitrary where T: Arbitrary` would be very useful for cutting down on a bunch of implementation boilerplate.

Edit: currently this fails clippy, but it seems like the failing code is unrelated to my diff, and is probably because of Rust 1.83 changing lint checks for elidable lifetimes.

### DIFF
--- a/src/foreign/core/cmp.rs
+++ b/src/foreign/core/cmp.rs
@@ -1,0 +1,23 @@
+use {
+    crate::{size_hint, Arbitrary, Result, Unstructured},
+    core::cmp::Reverse,
+};
+
+impl<'a, A> Arbitrary<'a> for Reverse<A>
+where
+    A: Arbitrary<'a>,
+{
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

fitzgen (approved):

fitzgen:
Looks like CI is failing due to an unnecessary lifetime: https://github.com/rust-fuzz/arbitrary/actions/runs/12189632357/job/34005838391?pr=210#step:5:30

fitzgen:
We should probably opt into a specific set of clippy lints, rather than whatever the defaults are, so that as new lints get added in the future our CI doesn't break from under us.

fitzgen:
Clippy fix in https://github.com/rust-fuzz/arbitrary/pull/211

PR #214, opened 2025-03-09
Don't generate invalid non-zero integers

evanrelf:
For example, the instance for `NonZeroUsize` was generating code equivalent to this:

```rust
NonZeroUsize::new(u.arbitrary::<usize>()?)?
```

But `u.arbitrary::<usize>()` can generate 0, which will fail.

### DIFF
--- a/src/foreign/core/num.rs
+++ b/src/foreign/core/num.rs
@@ -94,7 +94,7 @@ macro_rules! implement_nonzero_int {
     ($nonzero:ty, $int:ty) => {
         impl<'a> Arbitrary<'a> for $nonzero {
             fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
-                match Self::new(<$int as Arbitrary<'a>>::arbitrary(u)?) {
+                match Self::new(<$int as Arbitrary<'a>>::arbitrary(u)?.saturating_add(1)) {
                     Some(n) => Ok(n),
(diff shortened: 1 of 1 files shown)


src/foreign/core/num.rs
```diff
@@ -94,7 +94,7 @@ macro_rules! implement_nonzero_int {
     ($nonzero:ty, $int:ty) => {
         impl<'a> Arbitrary<'a> for $nonzero {
             fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
-                match Self::new(<$int as Arbitrary<'a>>::arbitrary(u)?) {
+                match Self::new(<$int as Arbitrary<'a>>::arbitrary(u)?.saturating_add(1)) {
                     Some(n) => Ok(n),
                     None => Err(Error::IncorrectFormat),
```
fitzgen, line 99: It is kind of nice that `1` becomes `NonZero(1)`, etc...

How about we change the `None => ...` arm to return `Self::new($int::MAX).unwrap()`?

evanrelf: I don't think it makes any difference. But I guess there's some intangible niceness about that. I can change it.

fitzgen (approved):
Thanks! Sorry for the delay, been traveling

evanrelf:
No worries! Hope your travels went well 🙂

Issue #217, opened 2025-04-08
List of breaking changes for next breaking release


fitzgen:
* [ ] https://github.com/rust-fuzz/arbitrary/pull/216
* [ ] remove split between `try_size_hint` and `size_hint`
* [ ] bump MSRV to latest stable

Anything else?

kpreid:
I was just working on improving some `Arbitrary` implementations, and I think un-splitting `try_size_hint` would be very much worthwhile to reduce the number of possible mistakes.

Other suggestions:

* #201 — add a `SizeHint` struct to make it much less verbose to construct size hints, since they could be combined using `+` and `|` operators (instead of methods).
* Consider flipping the definition of `depth` so that instead of accepting the current depth with a hard-coded `MAX_DEPTH`, the parameter is itself a maximum further depth, and recursion decrements rather than increments it. That way, if an application finds itself hitting a depth limit it can just ask for more depth instead of having to modify the library.

kpreid:
P.S. I'd be happy to try writing a PR for the `SizeHint` breaking change if there is interest.

fitzgen:
> [Consider a custom type for easier size hints #201](https://github.com/rust-fuzz/arbitrary/issues/201) — add a `SizeHint` struct to make it much less verbose to construct size hints, since they could be combined using `+` and `|` operators (instead of methods).

Sounds good to me. I'm fine with the operator overloading sugar, but I think there should still be named methods that the operators just forward to for better rust docs discoverability.

> Consider flipping the definition of `depth` so that instead of accepting the current depth with a hard-coded `MAX_DEPTH`, the parameter is itself a maximum further depth, and recursion decrements rather than increments it. That way, if an application finds itself hitting a depth limit it can just ask for more depth instead of having to modify the library.

Also sounds good to me 👍 

> P.S. I'd be happy to try writing a PR for the `SizeHint` breaking change if there is interest.

For just the types or the trait methods as well? Either way, that would be very welcome!

kpreid:
PR #218 adds the `SizeHint` type and combines `try_size_hint()` into `size_hint()`. (It does not change the meaning of the `depth` parameter.)

PR #218, opened 2025-04-10
Replace `(usize, Option<usize>)` with custom type `SizeHint` and eliminate `try_size_hint()`.

### LINKED ISSUE #201: Consider a custom type for easier size hints
When writing required manual `Arbitrary` implementations, one of the more painful elements is writing the `size_hint()` code, because it's a lot of `arbitrary::size_hint::*` function calls that must mirror the `arbitrary()` code. If `size_hint()` used a custom struct type instead of a tuple, this code could be made much more concise and legible using operators and method calls. For example, this `size_hint` method for one required and one possibly-required field,

```rust
fn size_hint(depth: usize) -> (usize, Option<usize>) {
    arbitrary::size_hint::and(
        Foo::size_hint(depth),
        arbitrary::size_hint::or((0, Some(0)), Bar::size_hint(depth)),
    )
}
```

could become something like

```rust
fn size_hint(depth: usize) -> arbitrary::SizeHint {
    Foo::size_hint(depth) + (arbitrary::SizeHint::ZERO | Bar::size_hint(depth))
}
```

(The original code could become shorter with some `use`s, but the `use` is itself additional boilerplate to write, particularly since it needs to be placed inside a `#[cfg(feature = "arbitrary")]` if there is one.)

As written, this is, of course, a breaking change of type, so it would require an `arbitrary` 2.0, and this isn’t worth that. I only wish to present this idea for consideration in case breaking changes are needed for other reasons.

kpreid:
The `SizeHint` type provides these advantages:

* Easier to construct by composition, using `+` and `|` or methods.
* More self-explanatory.
* Distinguishes between “unknown size” and “known to be unbounded” to aid debugging.

Since this is a breaking change to the trait, this PR also makes the closely related change of combining `try_size_hint()` and `size_hint()` into a single fallible function. This will make it easier to correctly implement the trait.

Also, while I was touching the macro code, I made all the crate paths absolute (`::arbitrary`), as they should be for the maximum compatibility.

Fixes #201.
Part of #217.

### DIFF
--- a/src/size_hint.rs
+++ b/src/size_hint.rs
@@ -1,149 +1,493 @@
-//! Utilities for working with and combining the results of
-//! [`Arbitrary::size_hint`][crate::Arbitrary::size_hint].
+//! The [`SizeHint`] type, for use with
+//! [`Arbitrary::size_hint()`].
 
-pub(crate) const MAX_DEPTH: usize = 20;
+use core::cell::Cell;
+use core::{fmt, ops};
 
-/// Protects against potential infinite recursion when calculating size hints
-/// due to indirect type recursion.
+use crate::Arbitrary;
+
+/// Bounds on the number of bytes an [`Arbitrary::arbitrary()`] implementation might consume.
 ///
-/// When the depth is not too deep, calls `f` with `depth + 1` to calculate the
-/// size hint.
-///
-/// Otherwise, returns the default size hint: `(0, None)`.
-///
-/// <div class="warning">This method is deprecated. Users should instead implement <a href="../trait.Arbitrary.html#method.try_size_hint"><code>try_size_hint</code></a> and use <a href="fn.try_recursion_guard.html"><code>try_recursion_guard</code></a></div>
-#[inline]
-#[deprecated(note = "use `try_recursion_guard` instead")]
-pub fn recursion_guard(
-    depth: usize,
-    f: impl FnOnce(usize) -> (usize, Option<usize>),
-) -> (usize, Option<usize>) {
-    if depth > MAX_DEPTH {
-        (0, None)
-    } else {
-        f(depth + 1)
+/// A size hint consists of a finite lower bound and a possibly infinite upper bound.
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
+pub struct SizeHint {
+    /// Inclusive lower bound.
+    lower: usize,
+
+    /// Exclusive upper bound. [`None`] means unbounded.
+    ///
+    /// Also contains some meta-information about the size hint as a whole.
+    upper: Upper,
+
+    out_of_fuel: bool,
+}
+
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
+enum Upper {
+    Bounded(usize),
+    /// The size is known to be unbounded (or overflow [`usize`]).
+    Unbounded,
+    /// The size is unknown either because a more specific size hint was not given,
+    /// or because computation of the size hint was cancelled before it completed.
+    /// This is functionally identical to `Unbounded` but prints differently.
+    ///
+    /// In these cases, the lower bound may also be lower than it should be.
+    Unknown,
+}
+
+impl SizeHint {
+    /// Equal to [`SizeHint::exactly(0)`][Self::exactly].
+    pub const ZERO: Self = Self {
+        lower: 0,
+        upper: Upper::Bounded(0),
+        out_of_fuel: false,
+    };
+
+    /// Indicates that a size hint is not available.
+    /// Returned by the default implementation of [`Arbitrary::size_hint()`].
+    ///
+    /// This is the most uninformative bound: zero to infinity bytes.
+    /// In addition, it prints differently from [`SizeHint::at_least(0)`][SizeHint::at_least],
+    /// to indicate that it is unknown rather than an unbounded collection.
+    pub const UNKNOWN: Self = Self {
+        lower: 0,
+        upper: Upper::Unknown,
+        out_of_fuel: false,
+    };
+
+    /// Indicates that the [`Context`] exhausted its `fuel`.
+    pub const OUT_OF_FUEL: Self = Self {
+        lower: 0,
+        upper: Upper::Unbounded,
+        out_of_fuel: true,
+    };
+
+    /// Creates a [`SizeHint`] for exactly `size` bytes.
+    #[must_use]
+    pub const fn exactly(size: usize) -> Self {
+        Self {
+            lower: size,
+            upper: Upper::Bounded(size),
+            out_of_fuel: false,
+        }
+    }
+
+    /// Creates a [`SizeHint`] for `size` bytes or more.
+    #[must_use]
+    pub const fn at_least(size: usize) -> Self {
+        Self {
+            lower: size,
+            upper: Upper::Unbounded,
+            out_of_fuel: false,
+        }
+    }
+
+    /// Returns the lower bound on bytes that will be consumed.
+    pub const fn lower_bound(self) -> usize {
+        self.lower
+    }
+
+    /// Returns the upper bound on bytes that will be consumed.
+    ///
+    /// Returns [`None`] if unlimited bytes could be consumed,
+    /// the number of bytes exceeds [`usize::MAX`],
+    /// or the bound is unknown.
+    pub const fn upper_bound(self) -> Option<usize> {
+        match self.upper {
+            Upper::Bounded(size) => Some(size),
+            Upper::Unbounded => None,
+            Upper::Unknown => None,
+        }
+    }
+
+    /// Take the sum of the `self` and `rhs` size hints.
+    ///
+    /// Use this when an [`Arbitrary::arbitrary()`] implementation is going to consume both
+    /// `self` bytes and `rhs` bytes, such as for the fields of a `struct`.
+    ///
+    /// This operation can also be written as the `+` operator.
+    #[inline]
+    pub const fn and(self, rhs: Self) -> Self {
+        Self {
+            lower: self.lower.saturating_add(rhs.lower),
+            upper: match (self.upper, rhs.upper) {
+                // checked_add causes overflow to become infinite
+                (Upper::Bounded(lhs), Upper::Bounded(rhs)) => match lhs.checked_add(rhs) {
+                    Some(upper) => Upper::Bounded(upper),
+                    None => Upper::Unbounded,
+                },
+                // If either is explicitly unbounded, then we can accurately say the result is.
+                (Upper::Unbounded, _) | (_, Upper::Unbounded) => Upper::Unbounded,
+                // If one is unknown and neither is unbounded, result is unknown.
+                (Upper::Unknown, _) | (_, Upper::Unknown) => Upper::Unknown,
+            },
+            out_of_fuel: self.out_of_fuel | rhs.out_of_fuel,
+        }
+    }
+
+    /// Take the sum of all of the given size hints.
+    ///
+    /// If `hints` is empty, returns `SizeHint::exactly(0)`, aka the size of consuming
+    /// nothing.
+    ///
+    /// Use this when an [`Arbitrary::arbitrary()`] implementation is going to consume as many
+    /// bytes as all of the hints together, such as for the fields of a `struct`.
+    #[inline]
+    pub fn and_all(hints: &[Self]) -> Self {
+        hints.iter().copied().fold(Self::ZERO, Self::and)
+    }
+
+    /// Take the minimum of the lower bounds and maximum of the upper bounds in the
+    /// `self` and `rhs` size hints.
+    ///
+    /// Use this when an [`Arbitrary::arbitrary()`] implementation is going to choose one
+    /// alternative and consume that many bytes, such as for the variants of an `enum`.
+    ///
+    /// This operation can also be written as the `|` operator.
+    #[inline]
+    pub const fn or(self, rhs: Self) -> Self {
+        Self {
+            lower: if self.lower < rhs.lower {
+                self.lower
+            } else {
+                rhs.lower
+            },
+            upper: match (self.upper, rhs.upper) {
+                (Upper::Bounded(lhs), Upper::Bounded(rhs)) => {
+                    Upper::Bounded(if lhs > rhs { lhs } else { rhs })
+                }
+                // If either is explicitly unbounded, then we can accurately say the result is.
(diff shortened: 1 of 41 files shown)


src/size_hint.rs
```diff
@@ -1,149 +1,379 @@
...
+    /// for generic types.
+    pub fn recursion_guard(
+        depth: usize,
+        f: impl FnOnce(usize) -> Result<Self, crate::MaxRecursionReached>,
+    ) -> Result<Self, crate::MaxRecursionReached> {
+        if depth > MAX_DEPTH {
+            Err(crate::MaxRecursionReached {})
+        } else {
+            f(depth + 1)
+        }
```
fitzgen, line 105: Let's switch from `depth` to `fuel`, from incrementing to decrementing, and from `depth > MAX_DEPTH` to `fuel == 0`.

I think we will may also want to preserve the fuel count across `size_hint` and `recursion_guard` invocations such that if we are given `fuel = 100` that doesn't mean "process type trees of up to depth 100" but instead "call up to 100 `size_hint`/`recursion_guard` methods". The latter is a better overall limit on compute spent on size hints. The former isn't very useful because a depth of 10 already gets to binary trees of size 1024 but also isn't very deep at all for non-pathological/non-exponential cases.

This probably involves defining and threading a new context type through `size_hint` calls and removing public constructors for `SizeHint` and replacing them with methods on the context, something like

```rust
pub struct SizeHintContext {
    fuel: usize,
}

impl SizeHintContext {
    pub fn new(fuel: usize) -> Self {
        Self { fuel }
    }

    fn consume_fuel(&mut self) -> Result<()> {
        if *self.fuel == 0 {
            return Err(());
        }
        *self.fuel -= 1;
        Ok(()) 
    }
    
    pub fn exactly(&mut self, size: usize) -> Result<SizeHint> {
        self.consume_fuel()?;
        Ok(SizeHint {
            lower: size,
            upper: Upper::Bounded(size),
        })
    }

    pub fn at_least(&mut self, size: usize) -> Result<SizeHint> {
        self.consume_fuel()?;
        Ok(SizeHint {
            lower: size,
            upper: Upper::Unbounded,
        })
    }
    
    // Other `SizeHint` constructors and combinators...
}

// ...

pub trait Arbitrary {
    fn size_hint(context: &mut SizeHintContext) -> Result<SizeHint> {
        SizeHint {
            lower: 0,
            upper: Upper::Unbounded,
        }
    }
}
```

It might even make sense to actually put the `SizeHint` inside the `SizeHintContext` and just have methods that mutate the size hint being built, rather than returning new size hints that we then need to combine. This would further disincentivize constructing a new `SizeHintBuilder` in the middle of a `size_hint` method implementation, which would defeat the purpose of having fuel in the first place. Doing it this way would also allow extracting the half-built `SizeHint` and getting at least an idea of the lower bound, even when we run out of fuel. (Maybe the `consume_fuel` method should set upper to unbounded if it returns an error?)

This approach effectively turns the context into a builder:

```rust
pub struct SizeHintBuilder {
    fuel: usize,
    lower: usize,
    upper: Upper,
}

impl SizeHintBuilder {
    pub fn new(fuel: usize) -> Self {
        Self {
            fuel,
            lower: 0,
            upper: Upper::Bounded(0),
        }
    }
    
    fn consume_fuel(&mut self) -> Result<()> {
        if *self.fuel == 0 {
            return Err(crate::Error::SizeHintOutOfFuel);
        }
        *self.fuel -= 1;
        Ok(()) 
    }
    
    pub fn finish(self) -> SizeHint {
        SizeHint {
            lower: self.lower,
            upper: self.upper,
        }
    }
    
    pub fn exactly(&mut self, size: usize) -> Result<()> {
        self.consume_fuel()?;
        self.lower = self.lower.checked_add(size).ok_or(())?
        self.upper += Upper::Bounded(size);
        Ok(())
    }

    pub fn at_least(&mut self, size: usize) -> Result<()> {
        self.consume_fuel()?;
        self.lower = self.lower.checked_add(size).ok_or(())?
        self.upper += Upper::Unbounded;
        Ok(())
    }
    
    pub fn unknown(&mut self) -> Result<()> {
        self.consume_fuel()?;
        self.upper += Upper::Unknown;
        Ok(())
    }
    
    // Other `SizeHint` constructors and combinators...
}

pub trait Arbitrary {
    fn size_hint(builder: &mut SizeHintBuilder) -> Result<()> {
        builder.unknown()
    }
}
```

What do other `arbitrary` maintainers think of the above approaches? I'm kind of partial to the builder-style API myself.

cc @Manishearth @nagisa

src/size_hint.rs
```diff
@@ -1,149 +1,379 @@
...
+}
+
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
+enum Upper {
+    Bounded(usize),
+    /// The size is known to be unbounded (or overflow [`usize`]).
+    Unbounded,
+    /// The size is unknown because a more specific size hint was not given.
+    /// This is functionally identical to `Unbounded` but prints differently.
+    Unknown,
```
fitzgen, line 30: I'm not convinced that there is a meaningful difference between these variants. Just having a different `Debug` print doesn't really seem worth it to me.

src/size_hint.rs
fitzgen: I guess the `and_all` and `or_all` combinators get a little trickier with the builder style API rather than the context-style API. We would either need to create sub-builders, which makes it easy to defeat the purpose of a single shared `fuel` value, or we would have to do some nasty generics magic which makes the API impenetrable. Or some other funky API to thread the fuel through.

This is making me lean back towards the context-style approach.

kpreid: > I guess the and_all and or_all combinators get a little trickier with the builder style API

Even `or()` is difficult to do that way because you need to take the minimum of multiple independent sums.

That said, I think it would make sense to track fuel in, not a “builder” (makes one output value) but a “factory” (makes many); methods of the factory are constructors of `SizeHint` values (or error).

src/size_hint.rs
kpreid: I included this feature precisely because, if it had existed, it would have helped me troubleshoot a missing size hint.

src/size_hint.rs
fitzgen: > That said, I think it would make sense to track fuel in, not a “builder” (makes one output value) but a “factory” (makes many); methods of the factory are constructors of `SizeHint` values (or error).

This sounds like the context-style API above, no? If not, then I am curious what you are imagining that is different.

kpreid: Sorry, you are correct, I think.

kpreid: I'm getting back to working on this (sorry for the long delay). I think I have discovered an elegant further improvement: If the `Context` interposes and counts fuel in every recursive call (that is, each `size_hint()` implementation calls `context.get::<FieldType>()` to recurse, not `FieldType::size_hint(context)`), then there is no need to use `Result` and the `?` operator to handle short-circuiting, and if we do it this way

* `size_hint()`s which hit the computation limit can have a better chance of returning a *lower* bound with partial information, rather than only an error
* If we don't also count fuel in constructors, then “leaf” `size_hint()` methods don't use the context at all and will always return the correct answer even if out of fuel, aiding the previous point
* no top-level callers ever have to decide how to handle `Err` and perhaps do it incorrectly

kpreid: Done! Please take a look and tell me what you think.

fitzgen: Apologies for the delayed response, and thanks for your patience.

What I liked about the previous iteration of the design was that you couldn't construct a `SizeHint` without checking/consuming fuel, which meant that it was practically impossible to misuse the APIs in such a way that you accidentally do not consume fuel when you should, leading to unbounded size-hint computations. The new version of this PR does not seem to preserve this property, as far as I can tell, and requires that users take care to call `context.get::<Foo>()` for nested types rather than `Foo::size_hint(context)`. While the new version of this PR does allow constructing "leaf" size hints without any fuel, I think that benefit is marginal in comparison to preventing unbounded size-hint computations.

Regarding

> `size_hint()`s which hit the computation limit can have a better chance of returning a _lower_ bound with partial information, rather than only an error

and

> no top-level callers ever have to decide how to handle `Err` and perhaps do it incorrectly

We could bring these benefits into the previous design via:

* Making the out-of-fuel error carry the partial size hint as a payload:

  ```rust
  pub struct SizeHintError {
      partial_size_hint: SizeHint,
  }
  ```

  (This would have getters and what not.)

* Adding a helper method to `Context` that unwraps the `Result`'s `Ok`/`Err` and returns the (potentially partial) size hint without any `Result`:

  ```rust
  impl Context {
      pub fn size_hint<T>() -> SizeHint
      where
          T: Arbitrary,
      {
          T::size_hint(self).unwrap_or_else(|e| e.partial_size_hint)
      }
  }
  ```

kpreid: > you couldn't construct a SizeHint without checking/consuming fuel,

Something I’d like to be able to offer someday — though there is no provision in the current trait/macro for doing so in practice — is the possibility of defining size hints for finite structures _as constants_. This would mean that infinite recursions would automatically be caught at compile time by the const evaluation limit, and run time doesn't need to compute an essentially constant value every time. That world can be mixed with the `get()` world, but I don't think it can be mixed with the "all construction consumes fuel" world (unless there is *also* some kind of `ContextWithInfiniteFuel` for static use).

That said, I do see your point and I'll switch it over if you still object. However, I don't think it is necessary to go back to using `Result` to do so.

fitzgen: > Something I’d like to be able to offer someday — though there is no provision in the current trait/macro for doing so in practice — is the possibility of defining size hints for finite structures _as constants_. This would mean that infinite recursions would automatically be caught at compile time by the const evaluation limit, and run time doesn't need to compute an essentially constant value every time. That world can be mixed with the `get()` world, but I don't think it can be mixed with the "all construction consumes fuel" world (unless there is _also_ some kind of `ContextWithInfiniteFuel` for static use).

This would be nice, but I wouldn't bend over backwards for it. In practice, LLVM should be able to inline and const-fold away 99% of `size_hint` implementations, so the benefit is really just for people writing `size_hint`s by hand (since we can presumably always do the right thing in the derive macro). All that is to say, there is also value in having a simple, easy-to-understand-at-a-glance API and sometimes that can outweigh fitting things into the static type system. Basically the same reason why we don't always use branded-lifetimes tricks (similar to the `ghost-cell` crate) every time we index into slices, and accept that bounds checks will happen at runtime (modulo llvm optimizing them away), even though we could use a convoluted API to prove safety at compile time and which would allow us to omit the runtime bounds checks.

fitzgen: > However, I don't think it is necessary to go back to using `Result` to do so.

Ah, this is a good point: the helper function I wrote is basically `context.get::<T>()` in the new PR iteration. I guess the one thing that `Result`s allow for is easily determining whether the result of a `size_hint` call is partial or not, but given that you've put an `out_of_fuel: bool` field inside `SizeHint` that can still be easily exposed to users (most of whom likely will not care, so making it less in-your-face API-wise as a method on `SizeHint` that userc can choose to query, rather than a variant they are forced to match on before doing what they really want to do, is probably a good trade off).

fitzgen: > I still think that interposing on recursions only is cleaner, but ease of getting it right is the most important thing.

Yeah, to be clear avoiding fuel on leaf types would be nice, but I think preventing footguns is definitely the more important property here.

src/size_hint.rs
```diff
@@ -1,149 +1,493 @@
...
+        if self.consume_fuel() {
+            T::size_hint(self)
+        } else {
+            SizeHint::OUT_OF_FUEL
+        }
+    }
+
+    /// Consumes 1 unit of fuel if available, and returns whether that fuel was available.
+    #[inline(never)]
+    fn consume_fuel(&self) -> bool {
```
fitzgen, line 318: Is there a reason you made this `inline(never)`? I would expect that we would want to make it `inline` instead, so that LLVM can fully inline and const-fold away `size_hint` implementations for small, non-recursive types.

src/size_hint.rs
```diff
@@ -1,149 +1,493 @@
...
+
+    /// Returns [`T::size_hint()`][Arbitrary::size_hint], and consumes 1 unit of fuel too.
+    #[inline]
+    pub fn get<'a, T: Arbitrary<'a>>(&self) -> SizeHint {
+        if self.consume_fuel() {
+            T::size_hint(self)
+        } else {
+            SizeHint::OUT_OF_FUEL
+        }
+    }
```
fitzgen, line 314: Two nitpicks:

1. I don't think we should make guarantees about how much fuel any particular method consumes. When describing what fuel is (say on the `Context::new` constructor), we can give rough estimates, but I think we want to keep it somewhat fuzzy so that we can adjust if necessary. The important part is that it provides a finite bound to size-hint computation, not *exactly* what that bound is.
2. As discussed in thread, we shouldn't actually need this method anymore, once `SizeHint` construction (even for leaf types) consumes fuel. Users can just call the trait methods directly instead.

src/size_hint.rs
```diff
@@ -1,149 +1,493 @@
...
+
+    /// Take the sum of the `self` and `rhs` size hints.
+    ///
+    /// Use this when an [`Arbitrary::arbitrary()`] implementation is going to consume both
+    /// `self` bytes and `rhs` bytes, such as for the fields of a `struct`.
+    ///
+    /// This operation can also be written as the `+` operator.
+    #[inline]
+    pub const fn and(self, rhs: Self) -> Self {
```
fitzgen, line 110: What do you think about moving all the and/and_all/or/or_all/repeat combinators to methods on `Context`? That way we retain the flexibility to make them consume fuel in the future, if that ever becomes necessary.

src/size_hint.rs
```diff
@@ -1,149 +1,493 @@
...
+pub struct SizeHint {
+    /// Inclusive lower bound.
+    lower: usize,
+
+    /// Exclusive upper bound. [`None`] means unbounded.
+    ///
+    /// Also contains some meta-information about the size hint as a whole.
+    upper: Upper,
+
+    out_of_fuel: bool,
```
fitzgen, line 22: I think we will want a method on `SizeHint`, and probably `Context` too, for checking whether we are out of fuel. I didn't see these below; apologies if these exist and I just missed it in my skim.

fitzgen (commented):
Thanks for your patience and sticking with this while we work out the design kinks!

I think we are now at a place where we agree what the design should look like (no `Result`s; `SizeHint` construction consumes fuel, even for leaf types) so if you update this PR accordingly, I can take a closer look and provide the hopefully final review.

Thanks again!

PR #238, opened 2026-06-02
Bring fix for fully qualified path

baloo:
When a downstream crate has a [minimal-version check], the version of derive_arbitrary that gets pulled in is 1.4.0, which does not carry the fix the fully qualified path on result that was made in de2becf7c6be9f916c9f1663c57bd3e4a84266a8

This bumps the patch version of the derive from arbitrary so that a downstream crate does not have to make a direct dependency expression against `derive_arbitrary`.

minimal-version checks are useful for checking API stability.

[minimal-version check]: <https://github.com/RustCrypto/actions/blob/master/.github/workflows/minimal-versions.yml>


Manishearth:
Generally I don't think it actually makes sense to do things like this: the version dep is "the latest version dep you need" which makes it maximally flexible for people. When you use `minimal-version` you are opting in to not getting fixes. If you need a fix, specify it in the Cargo.toml.

That said, that fix is years old now, so I don't particularly mind merging this. I don't plan to cut a release for this crate soon though (some other maintainer might).

PR #239, opened 2026-06-20
agent-driven: Fix soundness in try_create_array, add safety comments

Manishearth:
Found while running a safety audit with Gemini Next

The main safety bug was that we were creating an uninitialized slice type so we could write to it.

The safety comments are also agent generated, but they seem fine. I can remove them if desired.

### DIFF
--- a/src/foreign/alloc/ffi/c_str.rs
+++ b/src/foreign/alloc/ffi/c_str.rs
@@ -7,7 +7,12 @@ impl<'a> Arbitrary<'a> for CString {
     fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
         <Vec<u8> as Arbitrary>::arbitrary(u).map(|mut x| {
             x.retain(|&c| c != 0);
-            // SAFETY: all zero bytes have been removed
+            // SAFETY:
+            // Contract from `CString::from_vec_unchecked`: the vector must not contain
+            // any interior nul (zero) bytes.
+            // Evidence: `x.retain(|&c| c != 0)` removes all bytes equal to `0` from the
+            // vector `x`. Consequently, `x` contains no nul bytes, which guarantees
+            // it has no interior nul bytes.
             unsafe { Self::from_vec_unchecked(x) }
         })
     }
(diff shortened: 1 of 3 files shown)


src/foreign/alloc/ffi/c_str.rs
```diff
@@ -7,7 +7,12 @@ impl<'a> Arbitrary<'a> for CString {
     fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self> {
         <Vec<u8> as Arbitrary>::arbitrary(u).map(|mut x| {
             x.retain(|&c| c != 0);
-            // SAFETY: all zero bytes have been removed
+            // SAFETY:
+            // Contract from `CString::from_vec_unchecked`: the vector must not contain
```
nagisa, line 11: That's a neat way to write down safety comments in a more structured way. I will, um, yoink this.

If only there was a way to reference the contracts without repeating them a bunch from what's already in the doc for the method... 🤔

nagisa (approved):

src/foreign/alloc/ffi/c_str.rs
Manishearth: Yeah that's been my rule about good safety comments for a while: list the things you need to prove, then prove them.

Review becomes double checking you got all the invariants (easy if occasionally tedious) and then checking the local proof (usually easy, sometimes not)

src/foreign/alloc/ffi/c_str.rs
```diff
@@ -7,7 +7,12 @@ impl<'a> Arbitrary<'a> for CString {
...
+            // any interior nul (zero) bytes.
```
yancyribbens, line 12: Isn't the canonical name for the zero byte a `null` byte (not `nul` or `nil` for that matter) as per K&R =)

Manishearth: nul is what it is called in the context of strings

yancyribbens: thanks.  I had done a similar PR and debated to use `null` or `nul` when talking about an array of ascii bytes.  Sounds like I should have used `nul` in the comment.

